PR #127, opened 2025-10-31
Option to automatically go to new page on focus instead of clicking

### LINKED ISSUE #95: [FEA] - Investigate going to pages when focusing on nav drawer instead of clicking
### Description

Several users have commented that they wish just highlighting/focusing on a nav drawer item would load that page. This is the behavior of the Plex app.

Personally, I do not like this behavior, but since it seems like popular suggestion, I'll investigate maybe adding a toggle to enable it.

damontecres:
Closes #95 

Adds a setting (defaults to enabled) to automatically go to a page when the navigation drawer item is focused after a brief delay.

Basically, if you open the left nav drawer, and scroll down to "Movies" after a brief delay (600ms), the app will switch to the Movies page without having to click.

It can be disabled in advanced settings.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/preferences/AppPreference.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/preferences/AppPreference.kt
@@ -590,13 +590,25 @@ sealed interface AppPreference<T> {
                 getter = { },
                 setter = { prefs, _ -> prefs },
             )
+
+        val NavDrawerSwitchOnFocus =
+            AppSwitchPreference(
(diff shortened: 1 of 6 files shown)


Rand0mB0y:
Good idea, but personally, 600ms is too fast. Not even the time to read the labels. Also, perhaps this should be off by default? Otherwise, I'd consider something considerably slower like 1800ms or even more. 
Please keep up the good work @damontecres btw... your app is becoming better and better

damontecres:
I was mess with it more and I think you're right. I think this idea needs some more work.

PR #131, opened 2025-11-01
Use GetProgramsDto so that channelIds can be passed in the request body

joshjryan:


joshjryan:
Without paging, the Live TV screen is very slow to load with my 1500+ channels, but: this change allows it to load successfully without the 414.

image

damontecres:
Closing in favor of #132

PR #158, opened 2025-11-05
Keep border on current episode on series overview page

damontecres:
When moving off of the currently focused episode such as to go to the "More" menu, leave a slightly smaller and dimmer border on the episode's card.

This gives a visual hint on which episode is still selected.

unfocused


damontecres:
Decided against this because keeping the border might still make it look like it has focus at a glance. Will continue to workshop this.

PR #160, opened 2025-11-05
Optional setting for one click pause w/ d-pad center

### LINKED ISSUE #152: [FEA] - Optionally support 1-click pause for Android TV remotes?
### Description

Right now, if you use a standard Android TV remote (without play/pause buttons) you can only pause with 2 clicks where the first click brings up the player UI and the second click pauses.  We can already FF and RWD with 1-click, so can we also have the option to pause with 1-click? Perhaps with an onscreen pause icon (similar to the onscreen 30,60, etc icons) that goes away when we click again to play?

We can keep the current behavior as the default and 1-click as a possible option in the playback settings.

damontecres:
Closes #152 

Adds a setting to enable one-click-pausing via the D-Pad center button. It is disabled by default.

If enabled, when the video is playing, it will be pause and show the controls which is just like pressing the Pause or Play/Pause button.

If enable, when the video is paused and controls is not showing, it will resume playback including using the skip back on resume setting.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/preferences/AppPreference.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/preferences/AppPreference.kt
@@ -654,6 +654,18 @@ sealed interface AppPreference<T> {
                 summaryOn = R.string.enabled,
                 summaryOff = R.string.disabled,
             )
+
+        val OneClickPause =
+            AppSwitchPreference(
+                title = R.string.one_click_pause,
+                defaultValue = false,
+                getter = { it.playbackPreferences.oneClickPause },
+                setter = { prefs, value ->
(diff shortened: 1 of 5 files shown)


Pentaphon:
@damontecres as a follow up to this welcome feature, any chance you can NOT have the entire player UI come out when doing 1-click pause as requested in https://github.com/damontecres/Wholphin/discussions/990 ? 

Perhaps all that is needed is to just have a little pause icon toast popup that shows a pause or play icon at the bottom of the video when it is activated, similar to the 10 or 30 second circular icon that pops up when advancing left or right. Or maybe just not show anything at all, if you prefer true minimalism. Totally up to you. Whatever keeps the player UI from unnecessarily showing up is always welcome to keep the player from using too many resources.

Issue #85, opened 2025-10-28
[FEA] - Support external player


Pentaphon:
### Description

There's no option to use mpv within the app quite yet so in the meantime, can we have the option to use an external player?

Demo23019:
Could you add option like in dune play in external right from movie page

phamaralbr:
Being able to use an external player would be a big plus. I feel like the wholphin's player UI looks very dated, clunky and harder to use than the official jellyfin player.

husoyo:
This is by far the best jellyfin alternative. Desperately need external player option tho. I am waiting for the update.

Pentaphon:
> Desperately need external player option tho

I'm curious, which player do you use?

>I feel like the wholphin's player UI looks very dated, clunky and harder to use than the official jellyfin player.

There's issues to be improved but I much prefer Wholphin's player to the Official Android TV player.

husoyo:
I had Dune HD 8K but just sold it to get R_volution 4k mini. I haven't tried it too much yet but apart from being horrible at softwares side (encountered lots of annoying bugs), it gets the job done very well in terms of handling all kinds of big streams.

Pentaphon:
@husoyo I'm talking about what software player you use, not the hardware device you use.

husoyo:
> @husoyo I'm talking about what software player you use, not the hardware device you use. 

It's also called r_volution. The app is the player and it's UI. Look it up. I choose these devices because they hav VS10 Engine mainly

chasezgit:
100% need external player support. Jellyfin players are shit so I use vlc which is GREAT for any content, very capable player. Mpv works good but not for all hdr videos, exoplayer is absolutely useless.

reyaz006:
jellyfin-androidtv doesn't support external subtitles anymore so I'd love to check out any alternatives that provide such a support.

z3huti:
Please add option to play in  external player like in original Jellyfin or Moonfin

Issue #22, opened 2025-10-17
[BUG] - .ass subtitle not displayed properly


dzaqwan:
### Description

Stylized .ass subtitle sometimes too big, sometimes too small. The problem is more apparent for anime that have karaoke effect for the opening/ending songs where it appears to be glitched out

### App Version

v0.1.0

### Server Version

10.10.7

### Device

Fire TV Stick 4k Max

### Logs

```bash

```

damontecres:
Possibly related to #11. See also https://github.com/damontecres/Wholphin/issues/11#issuecomment-3411208619

damontecres:
The new `v0.3.0` release includes an experimental MPV backend which should display these subtitles correctly!

Pentaphon:
@damontecres any chance of using https://github.com/peerless2012/libass-android with Exoplayer so people that need Exoplayer for passthrough, etc can still get proper ASS support? 

Similar to how the official Android TV client is doing https://github.com/jellyfin/jellyfin-androidtv/pull/5121

damontecres:
> [@damontecres](https://github.com/damontecres) any chance of using https://github.com/peerless2012/libass-android with Exoplayer so people that need Exoplayer for passthrough, etc can still get proper ASS support?

Yup! Trying this out in #569

Issue #14, opened 2025-10-16
[FEA] VLC/MPV Support


danpergal:
### Description

Hello. First of all, I'd like to congratulate you and thank you for your effort in bringing us a new client for Android TV.
I wanted to ask if you are considering the possibility of adding another player like VLC or MPV in addition to Exo. Personally, I have had problems with Exo in many of my videos that VLC has been able to play without any issue. MPV also gives magnificent results playing formats that Exo struggles with.
Best regards and thank you very much.

damontecres:
Yes, I intend to investigate integrating alternative player backends. It's on the roadmap, but no ETA yet.

Pentaphon:
Please just focus on MPV instead of VLC.

damontecres:
The new `v0.3.0` release includes the experimental MPV backend. There is some follow up work though that is being tracked in #235.

Pentaphon:
@damontecres I think we can close this since VLC is redundant and MPV has been supported for over a month. Just trying to keep your issues clean, starting with the most requested ones.

damontecres:
MPV is added. No plans to add VLC since MPV is working very well.

PR #161, opened 2025-11-05
Add experimental MPV player backend

damontecres:
# Experimental MPV player backend

This is now mostly stable. It still needs further testing, but works well.

TODO:
- [x] Emit player events
- [x] Change tracks
- [x] External subtitles
- [x] Figure out better error access pattern after `release()`
- [x] Does CI work?
- [x] Update download progress indicator #165
- [x] Play next up
- [x] Playback speed
- [x] Double subtitles for ASS
- [x] Items might not be marked as watched even after dbb3701
- [x] ~Disable libmpv verbose logging for release builds~ will do in a follow up
- [x] Downloaded subtitles don't automatically activate
- [x] Use subtitle styling from #177
- [x] Live TV playback doesn't work
- [x] Loading circle shows at end of playback
- [ ] Shield upscaling doesn't recognize content - Not sure if this is fixable

Related to #14, #22, & #85

## Want to try it?

You can try this branch by following the [instructions here](https://github.com/damontecres/Wholphin/releases/tag/develop-mpv).

You can switch to MPV in advanced settings and toggle using hardware decoding or not.

## Other

This uses code and buildscripts from https://github.com/mpv-android/mpv-android, plus other third party libraries to build MPV

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,9 @@ This is not a fork of the [official client](https://github.com/jellyfin/jellyfin
 - Other (subjective) enhancements:
   - Subtly show playback position along the bottom of the screen while seeking w/ D-Pad
   - Force Continue Watching & Next Up TV episodes to use their Series posters
+- Different media playback engines, including:
+  - Default ExoPlayer/Media3
+  - Experimental MPV
 
 ### Roadmap
 


damontecres:
This is now pretty stable if folks want to try it.

Pentaphon:
> This is now pretty stable if folks want to try it.

@damontecres testing it now since I realized how easy it is to update to the MPV branch. My observations so far:

1) Hardware decoding performance is great. Videos actually start up faster, resume faster and skip ahead faster with MPV than Exoplayer on my Onn 4K Plus. The issue with wrong scaling on non-16:9 content with Exoplayer is non-existent. I can't find any playback issues on all my media. 

2) .srt subs are too big, .ass sub's are too small. I expect you to get requests for customizing the subs such as:

a) bold option (default is a bit Thin)
b) setting size of subs
c) setting sub offset from bottom of the screen 

That's about it so far! Great work! Will keep testing!

damontecres:
> srt subs are too big, .ass sub's are too small. I expect you to get requests for customizing the subs such as:

Once I've finished up with #177, I'll port it over to this branch too.

I do not have very many ASS/SSA subtitled files, so feedback for that is greatly appreciated!

Pentaphon:
@damontecres ok I finally found some issues while stress testing the player

1) In the Player UI, you can press the "next track" button with Exoplayer as much as you want to advance to the next video and it does so reliably and smoothly for 10, 20, 30 whatever videos, but when you do the same thing with mpv: 3 presses of the "next track" button while the UI is up will make the player UI "jump around" and become unresponsive until the player UI times out and closes itself. When it you pull the player UI back up, it is responsive again. 

2) I've had 4:3 videos show up as nothing but a black screen twice so far. You have to close out the player and start playing again to get the video to show up. I'm trying to capture a log for this but it's very random except it happened twice on 4:3 h264 videos. Could it be related to the API34 scaling bug you mentioned to me before since it affected only 2 4:3 videos so far? Possibly? No way to tell for sure.

3) I've had only 1 video not connect at all with mpv but connects fine with Exoplayer. I've logged myself using Exoplayer and playing the video just fine, then switching to mpv in the settings and trying to connect to the same video and getting a message saying "unable to get media URL from the server". Log below. 

[upload_Wholphin_0.2.7-41-g135bf90_20251108232110_cf4f21fde62044f28fcda35a933b0fc0.log](https://github.com/user-attachments/files/23435068/upload_Wholphin_0.2.7-41-g135bf90_20251108232110_cf4f21fde62044f28fcda35a933b0fc0.log)

damontecres:
The subtitle styling in MPV doesn't quite match 1:1 with #177. MPV seems to combine the text outline/shadow & background parameters (https://mpv.io/manual/stable/#options-sub-border-style) making them mutually exclusive.

Pentaphon:
> The subtitle styling in MPV doesn't quite match 1:1

That's probably fine. It might need some tweaking or 2 separate styling options in the future but I think the majority of people won't even notice. Most people just want the main styling options to be customizable such as font size, bold font, and outline. Shadow and background aren't as popular, from what I can see when people ask for subtitle options because most people don't notice shadow as much as they do outline and background takes up more space on the screen.

Pentaphon:
@damontecres the **bold option** in subtitle styling doesn't seem to apply itself to the mpv subs. Not sure if this has to do with what you mentioned earlier about 1:1 matching but this is one of the big ones people ask for.

Pentaphon:
@damontecres major issue found after actually watching a TV show! The video stops and the Google TV screensaver comes up after watching 10 mins of video on the MPV backend. Looks like the MPV playback doesn't suppress the screensaver from coming up as it does with Exoplayer. Does it every 10 mins like clockwork on my Onn 4K Plus. 

> The subtitle styling in MPV doesn't quite match 1:1

That's probably fine. It might need some tweaking or 2 separate styling options in the future but I think the majority of people won't even notice. Most people just want the main styling options to be customizable such as font size, bold font, and outline. Shadow and background aren't as popular, from what I can see when people ask for subtitle options because most people don't notice shadow as much as they do outline and background takes up more space on the screen.

GR3ee3N:
Dolby Vision and HDR10 files are tone mapped to SDR right now

MarcoCoreDuo:
> Dolby Vision and HDR10 files are tone mapped to SDR right now

MPV currently does not support native hdr playback on android, only tonemapping, like you noticed:
https://github.com/mpv-android/mpv-android/issues/874

Audio passthrough is broken too:
https://github.com/mpv-android/mpv-android/issues/619

For these features you need to use exoplayer.

damontecres:
@Pentaphon a bunch of things should be fixed:
1. Hitting next several times works
2. Player will flag the OS so the screensaver doesn't turn on during playback
3. Bold subtitles added

> 4:3 videos show up as nothing but a black screen twice so far.
> I've had only 1 video not connect at all with mpv but connects fine with Exoplayer

Not sure about these, but I did make some small changes to the device profile when using MPV, so might be worth trying again.

Edit: ~Needs to rebuild due to native code changes~ Done

VXsz:
- Pressing the menu button on the shield causes the player to quit (preventing me from checking some things like upscaling and screenshotting without hotkeys)

- Holding menu button on shield (to take screenshots without entering the menu thus quitting playback) successfully takes a screenshot but causes the app to freeze (only when the player UI is hidden).

- The app doesn't switch refresh rate and thus playback suffers from jitter (and even stutters), manually changing the refresh rate to 24000/1001 the content plays back nice, but there are some oceasonial stutters that happen, not sure why but it affects the official app and Dune as well.

- Pausing and resuming causes the player to go back to the specified amount in the settings, instead of it only happening when starting the episode (which would be the behavior I expected).



- Arabic subtitles suffer from font issues on the mpv player. In the vanilla mpv app (latest, august 25th build) it seems to display the subtitle fine with a default font (not the best choice I'd say but it's good), maybe mpv inside the app doesn't have a default/fallback font? Screenshots below. The one without boxes is the mpv app.








- Force transcoding for hi10p (h264 10bit), as most devices do not support it and mpv's SW decoder is too slow. Funnily enough there are many SoCs that do have hardware decoding for hi10p, but mpv doesn't recognize it (as they don't expose it as supported) and just falls back to software decoding, which is laggy and extremely slow. 2021-10-something builds of mpv had worse codec-support detection so you were able to use HW decoding for hi10p, but it had weird lag/random jumping. Maybe the jumping part is fixed on newer builds. If we can maybe find a way to enable HW decoding for these SoCs it would be cool (the shield and Roku devices for example support hw decoding hi10p).

Hope this helps. Also I think playback engine should be "auto" by default. Where it switches automatically to exoplayer for HDR and DV content.

Helpful Info:
Device: Shield TV Pro 2019
OS: Android TV, 9
App version is taken from this actions: https://github.com/damontecres/Wholphin/actions/runs/19236439703

Pentaphon:
@damontecres Thanks! Everything is fixed except: 
>1. Hitting next several times works

The behavior is different and somewhat worse this time.
a) On Exoplayer you can hit "next track" on the player UI as many times as you want and it will go through countless videos.
b) On mpv as of [v0.2.7-54-gef6998c](https://github.com/damontecres/Wholphin/releases/tag/develop-mpv) when you hit the "next track" button, you will go to the next video but then the focus goes to the scroll bar instead of staying on the "next track" button and you can't move to another UI element at all until the player UI closes instead of staying on the "next track" button. 

> 4:3 videos show up as nothing but a black screen twice so far.

I've yet to reproduce this but if I spot it again, I will make a dedicated issue for it with a log.

damontecres:
Thanks for testing, @VXsz!

> Pressing the menu button on the shield causes the player to quit

Fixed by #191 (and merged into this branch)

> Holding menu button on shield [...] causes the app to freeze

I haven't been able to reproduce this, but will keep trying.

> app doesn't switch refresh rate

No refresh rate switching is implemented for MPV yet. I have to investigate this.

> Pausing and resuming causes the player to go back to the specified amount

This is the intended behavior. But the setting summary might not be totally clear.

> Arabic subtitles suffer from font issues

The font I've included doesn't have Arabic gylphs which is causing this issue. `mpv-android` includes with a huge font file which does have many other languages. I switched away from that because its large and doesn't support bold, but I might need to revisit or work on other options.


> hit the "next track" button, you will go to the next video but then the focus goes to the scroll bar

@Pentaphon Should be fixed now

damontecres:
I've reverted back to use the same fonts as `mpv-android`. I have some ideas for the font issue, but I'll work on it in a follow up PR.

I'll have to follow up with refresh rate switching as well.

Otherwise this is feeling pretty stable I think.

VXsz:
>>  Holding menu button on shield [...] causes the app to freeze

>I haven't been able to reproduce this, but will keep trying.

I think this might've been a Shield thing. I started to notice it also happens on other apps. Although I'm not sure what's the root cause (might be an android tv thing).

Also something that I think would be cool is if we were able to edit the mpv.conf file, it's part of what makes mpv great (although it should be added with a warning when opened, to prevent misleading issue reports).

Oh in the track list (in the Dexter debug screenshot above), the external subtitle track lang is marked as null, when Jellyfin knows it is ara. I think it might be important to communicate with mpv/the player what the actual language should be. idk how much it actually affects playback but I think it would add more polish overall (like if one were to use the mentioned mpv.conf).

A little unrelated to this PR. I think the debug menu was slightly underwhelming. If it was more "verbose" (and compact) it would be a lot nicer. Take a look at Dune's debug/stats menu, I don't have a screenshot on me but it basically shows the server/client download speed in a nice graph while watching (usually bursts as expected but it also shows how healthy the network is) and it shows regular stream status information (direct play/remuxing/transcoding with specific details etc). Add on top some more stats perhaps and it would be a great way to share feedback from users, I think. mpv as well has a lot of frame-time stats which I think would be really great to have a look at (personally, at least).

btw, you are doing something big. Thanks a lot for taking on the effort and for what you've already done. Happy to test.

damontecres:
Thanks testing @Pentaphon & @VXsz!

Now that this PR is merged, you can switch back to stable or the [regular `develop` branch](https://github.com/damontecres/Wholphin/releases/tag/develop).

However, after you update the URL in the app settings, you'll see it say "No update available". You can press and hold on "No update available" and it will take you to the update page allowing you to force it back to the `develop` branch.

Also see #235 for follow up work on MPV

moi15moi:
@damontecres Any reason to not use https://github.com/jarnedemeulemeester/libmpv-android ?

Issue #92, opened 2025-10-28
[FEA] - Option to show TV or movie title logo in Player UI


Pentaphon:
### Description

It would be nice to see the movie or TV show logo in the player UI like we can in the official Jellyfin for Android client. Perhaps it can be optional instead of default for people who like the UI to be text-only?

Pentaphon:
@damontecres going through my issues and noticed this one can be closed.

PR #172, opened 2025-11-07
Add logo to playback overlay

damontecres:
## Details

Adds the logo (if available) to the upper left corner of the playback controls overlay.

I am still experimenting with how logos can be used on the details pages. It's problematic because of the variety of colors, eg dark logos can be hard to see on the dark background.

## Issues
Related to #92

## Screenshot
logo_playback Large


Pentaphon:
> It's problematic because of the variety of colors, eg dark logos can be hard to see on the dark background.

@damontecres The users can usually just change to a different logo using the Jellyfin server edit images function. At the very least, you can see how more visible that logo looks compared to the white text on those white lab coats. Perhaps the white text on the player can benefit from text outlining or a similar background you use for the buttons? Also, you probably won't need to put the title at the bottom anymore.

PR #192, opened 2025-11-11
Option to combine continue watching & next up rows

### LINKED ISSUE #19: [BUG] - Combining Up Next and Continuing Watching, puts the Up Next at the end of the row
### Description

Combining Up Next and Continuing Watching, puts the Up Next at the end of the row making it difficult to see/get too. Maybe you can sort it my dateplayed?

### Reproduction steps

Turn on combine up next and continue watching.
Complete and episode 
If you have more then a few Continue Watching shows/movies it becomes a chore to get to

### App Version

v0.1.0

### Server Version

10.10.7

### Device

ONN HD

### Logs

```bash

```

damontecres:
A real implementation for combining the `Continue Watching` and `Next Up` rows on the home page.

It requires a query for every next up item, so it could become slow especially for many items and larger servers. The app makes a best effort to cache the information.

The order is determined by querying each episode in next up to associate it to its immediately previous episode's last played date. Then both continue watching & next up items are reversed sorted by the last played data. So this only makes sense if you watch series in order.

Closes #19

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/services/DatePlayedService.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/services/DatePlayedService.kt
@@ -19,6 +19,7 @@ import org.jellyfin.sdk.model.api.request.GetEpisodesRequest
 import timber.log.Timber
 import java.time.LocalDateTime
 import java.util.UUID
+import java.util.concurrent.TimeUnit
 import javax.inject.Inject
 import javax.inject.Singleton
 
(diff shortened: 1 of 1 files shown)


damontecres:
On my server, with 25 item limit (the default), it takes an extra ~750ms to load. With 50 items, it takes 1.25s extra.

With caching, its <10ms extra, but can be inaccurate if uncommon situations. One scenario is if you watched Episode 8 a while ago this means E9 is showing in Next Up. If you go to replay E8, E9's position won't update. It should move to the beginning, but the cached value will be from before replaying E8.

Maybe there is a reasonable way to invalidate when a series is played.

damontecres:
The last played timestamps are now cached. Whenever a TV episode is played or marked watched/unwatched, will fetch new dates. Plus the data expires after 2 hours in case you watch on another device.

Finally, refreshing the home page (while on the home page, click Home on the nav drawer) will always refresh the order.

I think this will make this a pretty reliable feature until the server natively supports it.

PR #232, opened 2025-11-16
Add dimming effect to highlight which episode is currenty selected

damontecres:
On the series overview page, when moving off of the episode row, such as to do the play/more buttons, the other episodes will now become dimmer to indicate which episode you're interacting with.

Also, when moving back to the page from the nav drawer, focus is restored to where it left from. This is useful when scrolling fast left on the season tabs and you overshoot onto the nav drawer, moving right once will always move back to the season tabs.

In this example screenshot, Episode 3 is selected, so the others are darker.
dim Large

This is a follow up from #158


Pentaphon:
@damontecres I was thinking of a way to suggest making it more obvious which episode is selected and this looks like a good solution. The only suggestion I have is that judging from the screenshot, the dimmed screenshots may be a bit too dim for some people to make out when looking for another episode to select. Perhaps you should try reducing the dimming effect by say, 25% or 33%?

Issue #91, opened 2025-10-28
Have you built a new theme that could be needed in the theme presets? Show it off here to get feedback!


Pentaphon:
We have a color customizer that could potentially be in the works here: https://github.com/damontecres/Wholphin/issues/950 

However, if you have a preset theme you built from scratch **_that the community needs_**, show it off here to get feedback from the community! 

Current theme presets:

1. OLED Black
2. Purple 
3. Orange
4. Green
5. Blue
6. Darker Blue

damontecres:
I actually built the themes using https://material-foundation.github.io/material-theme-builder/ and tweaked a few of the colors, particularly the "border" color since that is not generated.

There is actually a library for generating such a theme in-app on device: https://github.com/jordond/MaterialKolor

So, I think eventually, integrating that library and enabling users to generate their own theme in-app might be the best solution. With a few default "curated" ones to start with.

Pentaphon:
> So, I think eventually, integrating that library and enabling users to generate their own theme in-app might be the best solution. With a few default "curated" ones to start with.

That sounds nice but if that proves to be too much work, I would be fine with just adding more themes like OLED Black, Dark Grey, A bolder shade of Blue, and a Dark Coffee color to really round out the current themes.

nomenon:
+1 oled theme

Pentaphon:
Repurposing this issue to let people submit theme proposals instead of starting a new theme. @YogiBear12 if you think one of your themes is needed, I highly suggest you show it off here whenever you have any free time. 

To anybody else reading this: feel free to request new themes that you think Wolphin needs.

SergeiAK:
## Theme proposal: Signal — black surfaces, orange for attention, green for state

I have been running a ninth preset on a Shield Pro for the past week and it has settled down, so here it is as a proposal.

**The idea.** Colour carries one meaning each, consistently:

- **green** — state: watch progress, watched marker, switches that are on;
- **orange** — attention: focus, and the technical labels (quality, tracks, times);
- **white** — focus where orange would collide with green, i.e. the playback buttons and the selected drawer item.

Hence the name. It is less "one more hue" than a rule for where colour goes.

**Two things I changed that are not about taste.** Both surfaced while working out why the accent looked washed out, and they apply beyond this theme:

[...]

PR #266, opened 2025-11-20
Add two new themes

damontecres:
Adds "Bold Blue" and Black/OLED themes

Also tweaks Orange & Green a bit to make the borders stronger and backgrounds darker/less colored

## Bold Blue
theme_bold_blue

## Black
This uses a 100% black background, so it friendlier with OLED displays.
theme_black

Related to #91


Pentaphon:
Thanks! Those 2 dark themes are definitely ones that users will request. The current blue one was a bit too bright for me. The brighter, more distinct color around the selected items really helps them stand out. 

I know my original issue asking for a "color grid" was a tall order but I think you can please most people by having them vote on the next theme they want. The only other theme I can think of that would look nice is a "Mocha" or "Dark Coffee" Dark Brown theme, if you are looking for new themes to add in the future. 

shades-brown-vector-color-palette-600nw-1479197399

nomenon:
nice, waiting in develop build)

PR #279, opened 2025-11-21
Updates to details pages including videos' stream details display

### LINKED ISSUE #239: [FEA] - UI content summary icons
### Description

Hi! Love your client. Thank you for your tremendous efforts on this. I personally think the design and layout is superior to the official app, having said that, some elements in the official app are very nice. For instance, the icons for the type of file and rating of the content would fit very well with your minimal display of text/icons.

OFFICIAL APP: Image

YOUR APP: Image

This would give more space below and perhaps with formatting of the director to the left you could even add the film logo to the right like in the official client?

Btw not sure if it's a bug, but on a 4K remux file your app doesn't seem to recognize it's a 4K file and simply lists the full filename.

OFFICIAL APP:
Image

YOUR APP:
Image

damontecres:
## Details

This PR updates the detail pages for Movies & TV Shows.

- Increase backdrop image size & brightness (now matches home page)
- Increase max width for titles
- Decrease max width for detail text
- Switch to using a customized display labels for video, audio, & subtitle streams

Additional change for movie pages:

- Don't dim movie details text
- Move video stream details below play buttons
- Move first row up higher

### Labels

Instead of using the server provided `DisplayTitle`, create display labels in-app. In general, this shortens the text without losing significant detail. It is similar to the official app.

Some of the changes:
- Don't show "SDR", only "HDR"
- Simplify audio channel info, eg "Surround 5.1" => "5.1"
- Show chosen subtitle language
- Show chosen subtitle codec
- Show "(+#)" if there are additional audio or subtitle tracks
- Use gray background slightly tinted by the theme color

## Screenshots

This is just an example:

#### Before

movie_before

#### After

movie_after

#### Before

series_before

#### After

series_after

## Issues
Closes #239

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/components/DetailsBackdropImage.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/components/DetailsBackdropImage.kt
@@ -1,0 +1,50 @@
+package com.github.damontecres.wholphin.ui.components
+
+import androidx.compose.foundation.layout.fillMaxHeight
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.draw.alpha
+import androidx.compose.ui.draw.drawWithContent
+import androidx.compose.ui.graphics.Brush
+import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.layout.ContentScale
+import androidx.tv.material3.MaterialTheme
+import coil3.compose.AsyncImage
+import com.github.damontecres.wholphin.ui.isNotNullOrBlank
+
+@Composable
+fun DetailsBackdropImage(
+    backdropImageUrl: String?,
+    modifier: Modifier = Modifier,
+) {
+    if (backdropImageUrl.isNotNullOrBlank()) {
+        val gradientColor = MaterialTheme.colorScheme.background
+        AsyncImage(
+            model = backdropImageUrl,
+            contentDescription = null,
+            contentScale = ContentScale.Crop,
+            alignment = Alignment.TopEnd,
+            modifier =
+                modifier
+                    .fillMaxHeight(.85f)
+                    .alpha(.75f)
+                    .drawWithContent {
+                        drawContent()
+                        drawRect(
+                            Brush.verticalGradient(
+                                colors = listOf(Color.Transparent, gradientColor),
+                                startY = size.height * .5f,
+                            ),
+                        )
+                        drawRect(
+                            Brush.horizontalGradient(
+                                colors = listOf(Color.Transparent, gradientColor),
+                                endX = 0f,
+                                startX = size.width * .75f,
+                            ),
+                        )
+                    },
+        )
+    }
+}


Pentaphon:
@damontecres This is definitely an improvement over the original details but I have to say: I think your client needs less text and more iconography and graphics. It doesn't have to look like the official client's icons but I think it would make your client more visually interesting while making each menu less text heavy. Maybe you have this as a long term plan already but I thought I'd just say that your client is unique in that it has a better style than the official client and you should lean into that more with more icons to give users more distinct and organized information.

I'd go into more of the direction Plex went with icons, badges and text, but with your own spin, like using very light variations of the current theme colors instead of just monochrome icons, badges and text. 

plex-hiding-film-details-now-v0-pki0zb217iva1

damontecres:
This PR evolved quite a bit and I ended up making some changes to the overall detail pages as well.

Pentaphon:
> This PR evolved quite a bit and I ended up making some changes to the overall detail pages as well.

Looks good! Glad you went with the theme color-tinted badges instead of monochrome badges like Plex and the official Jellyfin client went with and it definitely fixes the over-reliance on text the original details page had.

ClappTV:
It feels a bit inconsistent that in the movies section the elements appear beneath the buttons, while in the series section they are aligned to the right.
For better UX consistency, it might be ideal to maintain the same layout and visual hierarchy across all screens

Hukuma1:
iconsv3 (crude example)

What if we move the icons to the field where all the info is for the film like running time?

This also allows us to move the play button etc a bit down so it's not as crowded between the text (directed by).

Pentaphon:
> It feels a bit inconsistent that in the movies section the elements appear beneath the buttons, while in the series section they are aligned to the right. For better UX consistency, it might be ideal to maintain the same layout and visual hierarchy across all screens

I thought so too but the use of the episode screens in TV shows was also an different element to be considered. If I had to change it, I would put the info badges above the TV and movie synopses to keep everything together. 

> What if we move the icons to the field where all the info is for the film like running time?
> 
> This also allows us to move the play button etc a bit down so it's not as crowded between the text (directed by).

I feel like this makes the field too long and extends too much into the TV or movie backdrop. 

@Hukuma1 have you tried moving the star rating next to the year, duration and parents rating and putting the codec info badges under all that?

damontecres:
I appreciate the feedback everyone!

> It feels a bit inconsistent that in the movies section the elements appear beneath the buttons, while in the series section they are aligned to the right.

I did try it with Movie page labels on the right side (this is how Plex does it), but I thought it looked very strange just floating with nothing above or below even after trying to tweak spacing.

> What if we move the icons to the field where all the info is for the film like running time?

I tried this and I also tried putting them down a line next to the rating. But I agree with Pentaphon here, they extend too far. Part of this PR was also making the backdrop a little more visible.

Pentaphon:
@damontecres have you tried moving the star rating next to the year, duration and parents rating so they are all in 1 line and putting the codec info badges under that line?

damontecres:
> @damontecres have you tried moving the star rating next to the year, duration and parents rating so they are all in 1 line and putting the codec info badges under that line?

I had not. But now that I've mocked it up, I like that layout a lot!

Also, I think moving the labels on the series page is a good idea because there might be some languages that have much longer text for the buttons which could overlap.

Side note: I know there's space at the bottom of the series page, but I'm planning to add people row there so I want leave a little space.

movie_moved Large
<br/>
<br/>
series_moved Large

damontecres:
The labels are very eye catching in this new position on a full size TV screen though, so I'm playing with making the backgrounds semi-transparent.

Pentaphon:
> I had not. But now that I've mocked it up, I like that layout a lot!

Me too! The key was definitely to free up the line taken up solely by the star rating. The codec info badges also split up the text, which makes the entire details page less of a pool of text and it puts iconography in 2 different places instead of grouping it all together. Thanks for trying it out because I think this works!

>The labels are very eye catching in this new position on a full size TV screen though, so I'm playing with making the backgrounds semi-transparent.

You might be on to something there. Feel free to show us different transparency values if you want our feedback on nailing down the right setting.

Hukuma1:
Yes, this looks great!

I still feel like the logo of the show/movie can be added. What do you think? Could we add it to the right of the title maybe?

damontecres:
Thanks again all! Opened #323 to continue this

PR #292, opened 2025-11-21
Fix sentence structure for skip segment buttons for german language

Sk4hnt42:
Added a function to check if german language is beeing used.
String output will be changed in order, as in german word by word translation won't work.

We use noun + verb like "Intro Überspringen".

So instead of "skip + segment"  --> "segment + skip" will be used.

This will look like:

image

And:

image


damontecres:
Thanks for contributing!

This issue should probably be fixed by:
1) Using [formatted strings](https://developer.android.com/guide/topics/resources/string-resource#formatting-strings), so updating the `R.string.skip` string and using it like `stringResource(R.string.skip, stringResource(segment.type.stringRes))`
2) Adding new strings for "Skip intro", "Skip outro", etc and including translations for each whole phrase

Option 2 is probably better, but requires more translation work

Let me know if you want to make this change or I can do it in a day or two.

Sk4hnt42:
> Thanks for contributing!
> 
> This issue should probably be fixed by:
> 
> 1. Using [formatted strings](https://developer.android.com/guide/topics/resources/string-resource#formatting-strings), so updating the `R.string.skip` string and using it like `stringResource(R.string.skip, stringResource(segment.type.stringRes))`
> 2. Adding new strings for "Skip intro", "Skip outro", etc and including translations for each whole phrase
> 
> Option 2 is probably better, but requires more translation work
> 
> Let me know if you want to make this change or I can do it in a day or two.

Hi, thank you!
I've removed my PR, as you might be able to do this change much more efficient than me. 
But yes, option 2 seems to be the best option for this.

PR #323, opened 2025-11-25
Move stream labels to top of description

### LINKED ISSUE #162: [FEA] - Add ratings to home, smart screen library and favourites view
### Description

It would be awesome to be able to see the star ratings while scrolling the libraries, and without having to open the actual items.
There should be enough space between the title and the summary when it comes to home page and libraries, similarly to how the stock ATV client (and Dune) do. 
Furthermore, it would be great to have the same top "at a glance" style (title + stars + summary) in the favourites view, too.
Thanks for your work, btw. Really enjoying this client!

damontecres:
Follow up to #279
Closes #162

Move the stream labels to the top under the dot-separated details. Also move the rating up to the dot-separated details line and add it to home page.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/components/DetailsBackdropImage.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/components/DetailsBackdropImage.kt
@@ -1,5 +1,6 @@
 package com.github.damontecres.wholphin.ui.components
 
+import androidx.compose.foundation.layout.BoxScope
 import androidx.compose.foundation.layout.fillMaxHeight
 import androidx.compose.runtime.Composable
 import androidx.compose.ui.Alignment
@@ -14,7 +15,7 @@ import coil3.compose.AsyncImage
 import com.github.damontecres.wholphin.ui.isNotNullOrBlank
 
 @Composable
-fun DetailsBackdropImage(
+fun BoxScope.DetailsBackdropImage(
     backdropImageUrl: String?,
     modifier: Modifier = Modifier,
 ) {
@@ -23,10 +24,11 @@ fun DetailsBackdropImage(
         AsyncImage(
             model = backdropImageUrl,
             contentDescription = null,
-            contentScale = ContentScale.Crop,
+            contentScale = ContentScale.Fit,
             alignment = Alignment.TopEnd,
             modifier =
                 modifier
+                    .align(Alignment.TopEnd)
                     .fillMaxHeight(.85f)
                     .alpha(.75f)
                     .drawWithContent {


damontecres:
Debating how much to dim the background of the labels though. I am leaning towards 66%

## 100% brightness
labels_100

## 75%
labels_75

## 66%
labels_66

## 50%
labels_50

ClappTV:
I think 66% looks perfect, although there isn’t much difference between 100% and 66%.

I also think the options should be spaced out a bit more. For example, I see the title very close to the top edge of the screen, the quality labels too close to the synopsis, and the buttons are also too near the synopsis. I think each element should have enough separation so everything is clearly distinguishable.

I also think that the boxes for the people are too large. They should be circular so they take up less space on the screen, look the same, and don’t draw so much attention.

An example in case it gives you some ideas of how Plex does it
image

Pentaphon:
> I think 66% looks perfect, although there isn’t much difference between 100% and 66%.

I actually think 50% looks good. That's where it starts to blend in with the backdrop but still looking distinct. I wouldn't go any higher than 50% for the labels but I'd compromise at 66%

>I see the title very close to the top edge of the screen

I agree with @iPleyGit the title is too close to the top of the screen but that can be dealt with in another issue. 

>the quality labels too close to the synopsis

I think it's fine where it is. The tagline separates the labels from the synopsis. In fact, @damontecres you may want to look into italicizing the tagline to give it more emphasis and variety to all the text on the page. 

>the buttons are also too near the synopsis 

I disagree. I think the "Directed by" does a good job of keeping the buttons apart from the synopsis and the buttons should be at the bottom of all the movie info. 

>I also think that the boxes for the people are too large. They should be circular so they take up less space on the screen, look the same, and don’t draw so much attention.

This is actually a good idea. It solves those issues while being more Plex-like. It also fits with the rounded buttons and user icon on the top left corner. Totally approve.

ClappTV:
Maybe circular buttons wouldn’t look bad either instead of oval ones.

And instead of the background image covering the whole screen, which is often covered by the text, maybe make it like the home page — a gradient with the image appearing on the right side of the screen.

Pentaphon:
> Maybe circular buttons wouldn’t look bad either instead of oval ones.

Certainly. I suppose they could still elongate and show text when selected. Would also save space.

damontecres:
> > I think 66% looks perfect, although there isn’t much difference between 100% and 66%.

> I actually think 50% looks good. 

The difference is more pronounced on TV versus a computer screen I think. Going with 50% after using it for some time on my TV.

> I see the title very close to the top edge of the screen

I moved it down

> the quality labels too close to the synopsis
> the buttons are also too near the synopsis

I adjusted the padding just a little bit in the header. And moved the people row down a bit.

> italicizing the tagline

I knew it was missing something! This is good.

> the boxes for the people are too large. They should be circular

I have gone back and forth on this. Using circle means cropping which can end up imperfect for some pictures.

> Maybe circular buttons wouldn’t look bad either instead of oval ones.

I agree 100%. Unfortunately, google's TV foundation library does not and forces a minimum width causing an oval shape. So I would have to either make them taller to round out, which I think would waste space or I'd have to write the buttons from scratch which is annoying. I think I will eventually bite the bullet because I think the buttons are too large in general.

## New Screenshot

Screenshot_20251125_140436

Pentaphon:
> New Screenshot

Looks great! Moving the title down really made a difference in the overall presentation. The labels look great on my phone screen at max brightness and italicizing the tagline really makes it more obvious its a tagline. I think it's ready to merge unless other people see anything I miss.

ClappTV:
Now it really is perfect! What I do notice is that the title ‘people’ is far to the left of the pictures of the people.

Pentaphon:
> What I do notice is that the title ‘people’ is far to the left of the pictures of the people.

Looks like an intentional indent style choice. I'm fine with it  either way.

damontecres:
> > What I do notice is that the title ‘people’ is far to the left of the pictures of the people.
> 
> Looks like an intentional indent style choice. I'm fine with it either way.

Yeah, this is the style I like. It also gives some space to for the card to expand into when focused.

Thanks again everyone!

Issue #235, opened 2025-11-16
[TASK] - MPV Followups


damontecres:
Task tracking for the follow ups after adding the experimental MPV backend in #161

- Better font support (eg bold, custom and/or system fonts?)
- Refresh rate switching (#62)
- Pass language information for external subtitles
- Investigate options for an optional user supplied `mpv.conf`

Pentaphon:
Thanks for putting so much focus on mpv, which is attracting many users to your client since many other clients have unfinished mpv support, neglected mpv support in need of maintenance or outright refuse to add mpv support. Could you please sticky this next to the known issues? It will help a lot of newbies see it and try to help with building out these features.

moi15moi:
> * Better font support (eg bold, custom and/or system fonts?)

Mpv use libass to render text. To use android system font, you need to build fontconfig.

Pentaphon:
To everybody reading this: please test the new libplacebo branch here https://github.com/damontecres/Wholphin/pull/1077 before it gets merged.

Nocifer:
I was just about to create a new feature request directly related to sub-issue #297 . I don't have time right now to test the new branch, so before I create a possibly moot feature request, could I ask something first?

Would this solve the issue where MPV tries to direct play everything under the sun because it has no notion of device capabilities and device profiles, which results in playback of codecs like Hi10p and AV1 being extremely slow/laggy/choppy on devices that aren't powerful enough to SW decode them?

The feature request I was going to suggest is to expand the 'Prefer MPV' feature beyond HDR to also include a user-curated list of video and audio codecs that will also prompt Wholphin to switch back to Exoplayer, which I believe would solve the issue completely.

daftblazer:
I also would like the feature to to prefer MPV except for AV1. For some reason on my Amlogic Android TV box, AV1 hardware decoding works fine with exoplayer, but has issues with mpv. Exoplayer doesn't display ASS subtitles correctly most of the time, so I'd prefer to use mpv when I can. Its just a pain to always switch based on what I watch

Issue #179, opened 2025-11-09
[FEA] - Add option for edge style thickness in subtitle styling page


Pentaphon:
### Description

The only thing I see missing from testing out the new subtitle styling page is the edge style thickness customization option, which is currently being requested from the official Jellyfin client at https://github.com/jellyfin/jellyfin-androidtv/issues/5008 for reference.

It would be great to customize the size (or thickness) of the edge style to make it more distinct from the background, if need be. Thanks.

damontecres:
The outline thickness is hardcoded in `androidx.media`/`ExoPlayer`: https://github.com/androidx/media/blob/1.8.0/libraries/ui/src/main/java/androidx/media3/ui/SubtitlePainter.java#L107-L108

This is adjustable for MPV though.

Pentaphon:
> This is adjustable for MPV though.

That's going to be my default anyway so thanks!

Issue #12, opened 2025-10-16
[FEA] - Subtitle delay


Generator:
### Description

Add options for subtitle delay (e.g. -100ms, -1000ms, +100ms, +1000ms ...) on video player controls

juan-abia:
This would be incredible!!

damontecres:
#470 adds this for MPV. Implementing this in ExoPlayer will be _significantly_ more challenging.

SamJongenelen:
Any chance we can get it on exoplayer too?
I've found their examples repository https://github.com/anandwana001/exoplayer-example?

PR #327, opened 2025-11-26
Support custom `mpv.conf`

### LINKED ISSUE #305: [FEA] - Add a way to edit mpv.conf
### Description

I think this client is the best available, it works perfectly on Firestick using MPV, but it would be great if there was a way to edit the mpv.conf file directly from the app.

damontecres:
Part of #235 
Closes #305

Adds a preference for specifying a custom `mpv.conf`

Also adds a preference to toggle between `vo=gpu` and `vo=gpu-next`, see https://github.com/mpv-player/mpv/wiki/GPU-Next-vs-GPU

There are some caveats:
* Multi-line text entry via a remote is rough
* [Track selection options](https://mpv.io/manual/stable/#track-selection) may not work as expected because Wholphin manages this itself
* Bad settings will crash the app hard

Dev note: This rewrites the text field inputs to use the [state based implementations](https://developer.android.com/develop/ui/compose/text/migrate-state-based)

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/preferences/AppPreferencesSerializer.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/preferences/AppPreferencesSerializer.kt
@@ -68,6 +68,7 @@ class AppPreferencesSerializer
                                         .apply {
                                             enableHardwareDecoding =
                                                 AppPreference.MpvHardwareDecoding.defaultValue
+                                            useGpuNext = AppPreference.MpvGpuNext.defaultValue
                                         }.build()
                             }.build()
                     homePagePreferences =


Pentaphon:
> * [Track selection options](https://mpv.io/manual/stable/#track-selection) may not work as expected because Wholphin manages this itself

I totally support this feature for advanced settings since mpv has a lot of options. but I do also hope we still get common remote-friendly settings like changing the sub position on the screen and changing the subtitle style thickness for both media3 and mpv in the future. Mostly stuff that already be changed on mpv using keyboard shortcuts like: 

mpbindings

damontecres:
> > * [Track selection options](https://mpv.io/manual/stable/#track-selection) may not work as expected because Wholphin manages this itself
> 
> I totally support this feature for advanced settings since mpv has a lot of options. but I do also hope we still get common remote-friendly settings like changing the sub position on the screen and changing the subtitle style thickness for both media3 and mpv in the future. Mostly stuff that already be changed on mpv using keyboard shortcuts like:

@Pentaphon I think down the road, I will create a way to have Wholphin be able to defer to `mpv.conf` for the track selection.

But what do you mean by getting common remote-friendly settings? Do you mean, you would like to see remote friendly ways to adjust these settings during playback? How does this relate to the keyboard shortcuts? Sorry, can you clarify a bit?

Pentaphon:
> Sorry, can you clarify a bit?

@damontecres Certainly. I just mean that I hope settings which can be changed easily with a remote control should go to their own setting in the settings page rather than have them all be relegated to the mpv.conf file. For example, in my issue https://github.com/damontecres/Wholphin/issues/179 we could change the subtitle outline thickness with `sub-border-size` in the mpv.conf file, but I'd rather that setting be exposed in the existing subtitle settings page to be easily adjusted with our remote control. The mpv.conf file should just be for advanced users, while basic and intermediate settings should have their own entry in the settings page. 

>Do you mean, you would like to see remote friendly ways to adjust these settings during playback? How does this relate to the keyboard shortcuts? 

I only mean that the MPV keyboard shortcuts should be adapted to the Player UI. Requests like https://github.com/damontecres/Wholphin/issues/12 should end up right next to the video speed controls in the player UI.

damontecres:
>  settings which can be changed easily with a remote control should go to their own setting in the settings page rather than have them all be relegated to the mpv.conf file

Yes, this will absolutely be the case! 

> mpv.conf file should just be for advanced users, while basic and intermediate settings should have their own entry in the settings page

Yes, I want to find a good balance between the average and power users for settings. Certainly visual changes should have a D-Pad accessible setting and the app should preview them. Having to go back and forth between editing `mpv.conf` and playback for something like `sub-border-size` would be a nightmare.

PR #337, opened 2025-11-28
Add default sort option for collections

### LINKED ISSUE #83: [BUG] - server Collection Order not respected.
### Description

Collection Order is just alphabetical by default, while reordering is a nice feature, when first opened it should adhere to the original order.

### Reproduction steps

1. Click on '...'
2. Scroll to '....'
3. See error

### App Version

V0.2.0

### Server Version

10.10.7

### Device

Onn 4k

### Logs

```bash

```

damontecres:
Adds a "default" sort option for collections/boxsets. This makes the app request no sort at all and leaves it up to the server to order the items.

Fixes #83


CDubs00:
Thank you again for this! Seems to be working Great!

Issue #204, opened 2025-11-12
[BUG] - large libraries have issue with connection timed out with jf 10.11.2


vortex91:
### Description

In shows section the Toop Rated Unwatched is erroring out.  Im guessing because I have a large library?

### Reproduction steps

1. Click on '...'
2. Scroll to '....'
3. See error

### App Version

2.8

### Server Version

10.11.2

### Device

onn 4k 

### Logs

```bash

```

damontecres:
Querying for the "Top Rated Unwatched" (and Suggestions sometimes) seems to be much slower on Jellyfin 10.11 vs 10.10 for TV shows.

Issue #120, opened 2025-10-30
[FEA] - Poster rendering could use some further optimization?


Pentaphon:
### Description

I notice that posters with a lot of detail or sharp large text get a bit garbled when compared to the official Android for TV client. Here's a picture of the same poster, taken with the same camera on the same screen side by side. Note the text quality in the following example. Can anything be done to improve poster quality without sacrificing app performance? 

![Image](https://github.com/user-attachments/assets/329690a1-3ead-4b1f-ab57-502c14a390b9)

Pentaphon:
@damontecres just testing a clean install of 0.3.3-7 and all images, especially posters, look much better so I'm closing this! Thanks for following up on this and making your client the best looking Android TV client!

PR #359, opened 2025-12-02
Optimizes recommended content & image caching

### LINKED ISSUE #330: [BUG] - Image loading takes a long time to appear.
### Description

When you open the home page or any suggestion category, many sections stay with the text ‘loading…’ for several seconds before the images appear. From what I've been able to investigate, I think the issue is that the ‘recommended’ section loads a lot of content; in ‘recently released’ you can scroll through hundreds and hundreds of images. Maybe limiting the number displayed at first and then loading more as you scroll to the right would provide a more satisfying experience.

Another option could be to lower the image quality a bit so that loading becomes faster and the app feels more responsive.

### Reproduction steps

_No response_

### App Version

v0.3.2

### Server Version

10.11.3

### Device

Nvidia Shield TV

### Logs

```bash

```

damontecres:
## Details
This PR contains a number of optimizations

### Images
Updates to request images with maximum sizes

The max sizes are estimates on the maximum number of pixels that will be actually displayed. This means the server is downscaling large images which means 1) better quality downscaling and 2) smaller images sent over the network and cached

This should mean slightly better quality images and more images that can be cached locally

### Paging
Reduces network calls when fetching paginated data, ie grid pages and some rows. This PR combines the total record count & first page queries which slightly reduces total time to load paginated data.

Additionally, on movies/tv show recommended pages, the rows now only fetch a limited number of items (same as number on home page). While they previously fetched using pagination, there is an overhead required to paginate requests, so this is slightly faster.

### Async recommended fetching
On movies/tv show recommended pages, the rows are now fetched completely separately from each other. Previously, the rows were requested separtely, but each was only added to the UI in order as each completed. Now they will add as the data arrived.

Note: Top unwatched TV shows is still very slow on `10.11.x`, see #204

### Image cache

The local, on-disk image cache size is increased from 100mb to 200mb. Combined with the smaller images, this covers about 1500-2000 poster images, but will vary by server.

The cache size is now also configurable in advanced settings from 25mb to 1000mb. Additionally, you can see how much of both the disk and memory caches are in use.

## Issues
Closes #330 
Related to #120

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/util/profile/Codec.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/util/profile/Codec.kt
@@ -67,6 +67,7 @@ object Codec {
         const val VP8 = "vp8"
         const val VP9 = "vp9"
         const val AV1 = "av1"
+        const val VC1 = "vc1"
     }
 
     object Subtitle {
--- a/app/src/main/java/com/github/damontecres/wholphin/util/profile/DeviceProfileUtils.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/util/profile/DeviceProfileUtils.kt
@@ -92,9 +92,11 @@ fun createDeviceProfile(
     val avcHigh10Level = mediaTest.getAVCHigh10Level()
     val supportsAV1 = mediaTest.supportsAV1()
     val supportsAV1Main10 = mediaTest.supportsAV1Main10()
+    val supportsVC1 = mediaTest.supportsVc1()
     val maxResolutionAVC = mediaTest.getMaxResolution(MimeTypes.VIDEO_H264)
     val maxResolutionHevc = mediaTest.getMaxResolution(MimeTypes.VIDEO_H265)
     val maxResolutionAV1 = mediaTest.getMaxResolution(MimeTypes.VIDEO_AV1)
+    val maxResolutionVC1 = mediaTest.getMaxResolution(MimeTypes.VIDEO_VC1)
 
     // / HDR capabilities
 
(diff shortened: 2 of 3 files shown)


damontecres:
This is in draft because any changes to how caching works should always require additional thought and testing! But my initial tests are good so far.

Pentaphon:
> This should mean slightly better quality images and more images that can be cached locally

Happy to see this included in this PR. Right now the poster images look a bit too garbled if the poster is too detailed or has serif fonts in the title so I am curious to see if you found a happy medium where posters look "good enough" but also use less space. I'd be happy to provide feedback before this hits the stable channel.

damontecres:
There is an improvement to posters I think. Here's an example, you can see the glow around the title is more clear

The original on the server is 1700x2550 pixels, 479kb. With this PR, it downscaled to ~267x400, 95kb.

## Before
before

## After
after

Pentaphon:
>There is an improvement to posters I think. Here's an example, you can see the glow around the title is more clear

>The original on the server is 1700x2550 pixels, 479kb. With this PR, it downscaled to ~267x400, 95kb.

Yep, that's what I was looking for. The rough, jagged edges on lines and fonts are not now smoothed out and that "dot matrix" look that appeared on faces and heavy detail is mostly gone. The posters don't have to look perfect but before this PR, they were sort of rough looking on a 4K TV. 

You might even be able to cut down the filesize even further depending on what downscaling methods you have available to you that don't affect performance. What was the downscaled filesize before this PR?

damontecres:
> You might even be able to cut down the filesize even further depending on what downscaling methods you have available to you that don't affect performance. What was the downscaled filesize before this PR?

Before this PR, images were downloaded (and cached) at full size. They would be scaled down as needed.

My understanding is that Android's default down scaling is optimized for speed only, so it doesn't do anti-aliasing for example causing the jaggedness. There are libraries and some tweaks to improve this, but since Jellyfin's server supports down scaling already (and seems to do so very well) and using that means Wholphin can cache more images, it makes sense to use it.

Issue #72, opened 2025-10-27
[FEA] - Additional types of listings


J-o-h-n-M:
### Description

Coming from KODI, I am used to having my series just in lists. So no posters or banners, just a list. 

Would be nice if we can have a few options to select or own preference for listing movies/series.

Spark-NV:
+1
a list view is something every client appears to be missing sadly. i feel a list view is so much cleaner and easier to navigate.

im hoping got a view like this:

![Image](https://github.com/user-attachments/assets/50b6f703-3ac2-493b-9057-8a3e7757b9e7)

PR #368, opened 2025-12-04
Customize display of library grid pages

### LINKED ISSUE #186: [FEA] - Ability to change view once inside collection to be similar to how home page is
### Description

One of the main reason I love this app is how I can scroll on items and view discription on top left in all the sections in home page and shows and movies etc... but in collections when I enter a particular collections it displays all items with out the top left discription.   Can we add this option (once inside collection) to view description on top left.

damontecres:
## Details

Adds options to customize how grid pages are displayed. Every grid page can be individually customized.

The options are saved per user on the device and restored when navigating back to the page. The pages that support it have a new button next to the sort/filter buttons and changes are updated live.

The options are:
* Primary or Thumb image
* Aspect ratio
* Show details header w/ backdrop (similar to the one on the home page)
* Number of columns
* Spacing between cards
* Card image content scale (eg Fit, Fill, Crop, etc)

Note: the Genre tabs' grids cannot be customized yet

### Other changes
- Nav drawer now opens over top of content instead of pushing it
- Images for most cards are fetched with actual fill size

### Possible future work
* Customize Genre grids
* Add more image types, banners?
* Add similar customization options to pages with rows, eg Home page and Recommended tabs

## Example

For example, this is a movie library using thumb images with 8 columns and reduced spacing. It also shows the details header

### Screenshot
grid_view_options

## Issues
Closes #186
Related to #72

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/cards/ItemCardImage.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/cards/ItemCardImage.kt
@@ -88,8 +88,8 @@ fun ItemCardImage(
         watchedPercent = watchedPercent,
         modifier =
             modifier.onLayoutRectChanged(
-                throttleMillis = 1000,
-                debounceMillis = 250,
+                throttleMillis = 100,
+                debounceMillis = 25,
             ) {
                 size = IntSize(width = it.width, height = it.height)
             },


damontecres:
~And I immediately found several bugs, so this is still a work in progress!~ Fixed

~Also fewer columns means larger images which can conflict with the new max sizes in #359, so I need to look at that~ Fixed

PR #381, opened 2025-12-06
Fix combining continue/next-up for new episode premieres

### LINKED ISSUE #377: [QST] - Combined Continue Watching and Up Next Sorting
### Question

Not sure I am correct, and if I am, if it’s possible to even implement. Continue Watching and Up Next seem to be sorted a little incorrectly. Continue Watching seems to work as expected ordered by last played datetime, but Up Next should use the earliest unwatched episodes release date OR the last played datetime whichever is greater. That way if a show was on break or it premiered a new season it’ll show up in the list towards the front. 

I have only been able to do limited testing though so I maybe missing something.

damontecres:
Better handling for combining continue/next-up by considering the premiere date of the episode. So episodes will use the previously watched one's last played date or the episode's premiere date, whichever is newer.

This means when a new season comes out or the show returns from a break, the new episodes will show earlier in the list like the next up rows shows when not combined.

Fixes #377

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/services/DatePlayedService.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/services/DatePlayedService.kt
@@ -8,7 +8,6 @@ import com.github.damontecres.wholphin.preferences.AppPreference
 import com.github.damontecres.wholphin.services.hilt.IoCoroutineScope
 import com.github.damontecres.wholphin.util.GetEpisodesRequestHandler
 import com.google.common.cache.CacheBuilder
-import com.google.common.cache.CacheLoader
 import dagger.hilt.android.qualifiers.ActivityContext
 import dagger.hilt.android.scopes.ActivityScoped
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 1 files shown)


CDubs00:
Quick clarification; does it use the latest episodes premiere date or the first unwatched episodes premiere date? If it’s the former that may cause issue of a weekly release to move to the front even if the show isn’t currently being watched and it’s not a premiering season if it’s the latter the premiering season will show up in the front but if the show goes unwatched it would move down the list.

damontecres:
> does it use the latest episodes premiere date or the first unwatched episodes premiere date

The first unwatched episode's premiere data. This is the episode that would show on the regular uncombined Next Up row.

I still don't think this is perfect, but not sure if there's a better way. I'm going to read through the server's code/docs to see if there's any good ideas there.

CDubs00:
> The first unwatched episode's premiere data. This is the episode that would show on the regular uncombined Next Up row.
> 
> I still don't think this is perfect, but not sure if there's a better way. I'm going to read through the server's code/docs to see if there's any good ideas there.

I think that will cover most cases with the best outcome.

PR #382, opened 2025-12-06
Automatic refresh rate switching setting

### LINKED ISSUE #62: [FEA] - Auto Frame Rate Switching and Auto Resolution Switching
### Description

First off all, many thanks for your hard work. 
I like the app a lot, way better than the official app that is ugly and in my opinion outdated.
I would like to ask if its possible to add to your roadmap, the auto frame switching and the auto resolution switching. 
Once again, thank you very much for your work.

### LINKED ISSUE #379: [FEA] - switch framerate
### Description

At first, I have to say that this app looks awesome and is already very good. There is just one thing that makes it unusable for me: I can’t stand pulldown stutter. So it would be great if there were a way to change the framerate automatically.

damontecres:
Adds an advanced playback setting to enable automatic refresh rate switching during playback

When enabled, the app looks at the available display modes, chooses the best candidate, and attempts to switch to it. After playback ends, the display is switched back to the original mode.

Closes #62
Closes #379

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/services/DatePlayedService.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/services/DatePlayedService.kt
@@ -8,7 +8,6 @@ import com.github.damontecres.wholphin.preferences.AppPreference
 import com.github.damontecres.wholphin.services.hilt.IoCoroutineScope
 import com.github.damontecres.wholphin.util.GetEpisodesRequestHandler
 import com.google.common.cache.CacheBuilder
-import com.google.common.cache.CacheLoader
 import dagger.hilt.android.qualifiers.ActivityContext
 import dagger.hilt.android.scopes.ActivityScoped
 import kotlinx.coroutines.CoroutineScope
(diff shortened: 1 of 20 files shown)


Pentaphon:
Looking forward to testing this on both Exoplayer and MPV. This is a crucial feature that we didn't have.

Pentaphon:
@damontecres just started to test the rate switching setting and one thing popped out at me. 

- When rate switching is enabled, Wholphin plays ALL content, whether its 720p or 1080p, in 2160p on a 4K TV. 
- Meanwhile, when rate switching is enabled: the official Android TV client plays 720p and 1080p content in 1080p on a 4K TV. 

Just thought I'd mention this here instead of making an issue. I'll continue testing. MPV seems to be faster than Exoplayer at rate switching from what I can see, which is pretty neat!

damontecres:
> the official Android TV client plays 720p and 1080p content in 1080p on a 4K TV

Yeah, the official app seems to combine refresh rate & resolution switching together. This PR solely deals with refresh rate and it is up the OS (really the user's TV & android settings) to pick the resolution just like regular playback.

The original issue #62 does mention resolution switching. I can work on that as a separate settings I guess.

Issue #441, opened 2025-12-12
Feature Request: Display end time of movies/TV episodes before playback


Bastien42:
Currently, Wholphin does not display the end time of movies or TV show episodes before playing the content. This information would be useful for users to quickly assess how much time they need to allocate for watching, especially when browsing their library.

**Proposed Solution:**
Display the end time next to the duration of the content, similar to how the native Jellyfin app handles it.

**Example:**
- "Duration: 2h 15m | Ends at: 10:30 PM"

**Benefits:**
- Provides at-a-glance scheduling information for users.
- Improves user experience by making it easier to plan viewing sessions.

**Additional Context:**
This feature would be particularly helpful for users who want to know when a movie or episode will finish before starting playback.

Bastien42:
I wanted to add that your app is awesome. I've finished the French translation, and I'm really impressed by how fast you came up with such a feature-rich app. I'm really glad you put the work into that and hope you'll maintain that as long as you have fun with it. Thanks!

Pentaphon:
@Bastien42 you can close this issue now. The PR author didn't use the command to close this PR once it was merged. Glad you got your request filled.

PR #475, opened 2025-12-15
Add "Ends at" time display.

asandhu3:
Addresses #441 .

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/components/MovieComponents.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/components/MovieComponents.kt
@@ -1,5 +1,6 @@
 package com.github.damontecres.wholphin.ui.components
 
+import android.content.Context
 import androidx.compose.runtime.Composable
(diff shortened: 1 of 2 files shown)


Pentaphon:
Is this going to be placed inside the play button?

damontecres:
#493 will unblock this, there's an example of how to use it in that PR

When you update this PR, can you provide some screenshots?

Thanks!

asandhu3:
@damontecres 
Screenshot 2025-12-17 at 2 20 19 PM
Screenshot 2025-12-17 at 2 20 37 PM
Screenshot 2025-12-17 at 2 21 18 PM

damontecres:
I consolidated the code for movies & episodes into a a single function.

I also moved the string to be after the time remaining time. And changed the calculation to use the time remaining or the whole duration if the item's time remaining is zero.

Finally, removed the end time from the series header since that duration is an estimate of the average runtime for episodes, not a specific playable runtime.

image


Thanks again for the contribution!

PR #481, opened 2025-12-16
Update MPV playback threading model

damontecres:
There's no user facing changes in this PR

This PR updates `MpvPlayer` to use a threading model more similar to [ExoPlayer's model](https://developer.android.com/reference/androidx/media3/exoplayer/ExoPlayer#threading-model).

There's now a dedicated thread looper message queue between the UI interactions on the main thread and calls to `libmpv`. This is necessary because previously, the main thread would make the calls into `libmpv`. `libmpv` uses internal mutex locking for some operations which meant that the main thread would be blocked momentarily.

Now, the UI issues a command (eg seek to X position) and the command is added to the thread looper queue and processed on that separate thread. This uses the standard Android Handler/Looper mechanisms.

Related to #235

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/services/AppUpgradeHandler.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/services/AppUpgradeHandler.kt
@@ -188,7 +188,7 @@ suspend fun upgradeApp(
         }
     }
 
-    if (previous.isEqualOrBefore(Version.fromString("0.3.6-11-g0"))) {
+    if (previous.isEqualOrBefore(Version.fromString("0.3.6-52-g0"))) {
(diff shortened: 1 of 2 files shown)


damontecres:
This also fixes a bug where using the `gpu-next` toggle from #327 wasn't actually being applied. Switching on `gpu-next` on my Shield just shows a purple screen with audio.

Need to investigate more on other devices

damontecres:
So far, of my test devices, only the NVIDIA shields have a problem with `gpu-next`. GCwGTV, Onn, Fire TV all are fine.

The shield reports OpenGL errors, which I suppose isn't that surprising because I know there are other issues with OpenGL on them.

Since they're pretty popular, the latest commit will turn off `gpu-next` for Shields when the update is installed. Users can still turn it back on if they so wish.

Issue #485, opened 2025-12-16
[FEA] Add text outlines or drop shadows to OSD overlay


zoro11031:
### Description

**Discussion**

The OSD text during playback (title, episode info, playback time) is rendered as pure white with no outline or shadow. This makes it difficult or impossible to read when video content has light backgrounds - especially on a TV viewed from couch distance.

**Example**

See attached photo (taken from couch distance) - episode metadata is barely legible against the bathroom tiles in the scene.

![Image](https://github.com/user-attachments/assets/a06b6ea6-9f38-4e64-822a-c70dbafe9b22)

**Suggested Solution**

Add a dark stroke/outline (1-2px) or drop shadow to all OSD text elements. This is a standard approach used by VLC, mpv, Plex, and most video players to ensure readability regardless of video content.

Alternatively, consider a semi-transparent dark gradient behind the text overlay area (similar to YouTube's approach).

**Affected Elements**

- Title/episode name
- Playback timestamp
- Any other text rendered over video

jesse-kaufman:
I agree. A translucent black background would make everything in the OSD easier to read

PR #486, opened 2025-12-17
Add text shadow to playback overlay for better readability

zoro11031:
## Description

Add OsdTextShadow to all text elements in the playback overlay UI to ensure readability against light video content. The shadow uses a semi-transparent black color with a slight offset and blur.

Applied to:
- Time display
- Seek bar position/remaining time
- Chapter and queue titles
- Controller title/subtitle
- End time display
- Skip indicator

### Related issues

Solves: https://github.com/damontecres/Wholphin/issues/485

### Screenshots

new_font

text_shadow_screenshot_2

### AI/LLM usage
This change was written entirely using Claude Code, though the change itself is small. Tested using Android TV emulator in Android Studio.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/UiConstants.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/UiConstants.kt
@@ -41,11 +41,12 @@ sealed class AppColors private constructor() {
 /**
  * Text shadow for OSD/playback overlay text to ensure readability against light video content
  */
-val OsdTextShadow = Shadow(
-    color = Color.Black.copy(alpha = 0.8f),
-    offset = Offset(1f, 1f),
-    blurRadius = 3f,
-)
+val OsdTextShadow =
+    Shadow(
+        color = Color.Black.copy(alpha = 0.8f),
+        offset = Offset(1f, 1f),
+        blurRadius = 3f,
+    )
 
 const val DEFAULT_PAGE_SIZE = 100
 
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackOverlay.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackOverlay.kt
@@ -53,8 +53,8 @@ import com.github.damontecres.wholphin.data.model.Playlist
 import com.github.damontecres.wholphin.data.model.aspectRatioFloat
 import com.github.damontecres.wholphin.ui.AppColors
 import com.github.damontecres.wholphin.ui.AspectRatios
-import com.github.damontecres.wholphin.ui.OsdTextShadow
 import com.github.damontecres.wholphin.ui.LocalImageUrlService
+import com.github.damontecres.wholphin.ui.OsdTextShadow
 import com.github.damontecres.wholphin.ui.TimeFormatter
 import com.github.damontecres.wholphin.ui.cards.ChapterCard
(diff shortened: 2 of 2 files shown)


Pentaphon:
I was about to request an outline but this works well enough.

High-rider:
I think a gradient shadow would be better as it'll eliminate that straight shadow line from the top of the overlay. Bottom area can be a bit darker than the top.

Pentaphon:
> I think a gradient shadow would be better as it'll eliminate that straight shadow like from the top of the overlay

@zoro11031 can you confirm?

zoro11031:
> I think a gradient shadow would be better as it'll eliminate that straight shadow line from the top of the overlay. Bottom area can be a bit darker than the top.

I can mock it up with a gradient tonight after work  to see how that looks. Can you be elaborate on where I need to look to see the “ straight shadow line from the top of the overlay” that you’re referring to? I don’t think I’m seeing what you’re seeing, but I am colorblind and sometimes have trouble with picking out stuff like that so it might just be my eyes haha.

@High-rider

High-rider:
@zoro11031  This line
527365116-f94230cd-96b1-437f-bbc1-30f5ef0ff788

For gradient, you can try keep the top part lighter and the bottom part darker. also you can test various lengths of as to where this background overlay shadow starts, like currently in this, it starts from slightly above of the control buttons, you can try various settings with gradient, try from the text where show/movie name is written which goes till the bottom.

2nd option, this type of background shadow appears on whole screen when the controls are out to better highlight them.

You can try and see which one looks good, if possible paste screenshots here as well

zoro11031:
> @zoro11031 This line 527365116-f94230cd-96b1-437f-bbc1-30f5ef0ff788
> 
> For gradient, you can try keep the top part lighter and the bottom part darker. also you can test various lengths of as to where this background overlay shadow starts, like currently in this, it starts from slightly above of the control buttons, you can try various settings with gradient, try from the text where show/movie name is written which goes till the bottom.
> 
> 2nd option, this type of background shadow appears on whole screen when the controls are out to better highlight them.
> 
> You can try and see which one looks good, if possible paste screenshots here as well

@High-rider I think what you're looking at is just a shadow on the ground in the scene in question. If you look at my second screenshot below it showcasing a different scene you'll see what I mean. My change is only adding a subtle shadow to the text elements themselves and not a dark gradient behind them.

I can add a dark background gradient though. Not sure if that would be redundant with the text shadows though. I initially was considering doing one or the other but both might be best?

High-rider:
Okay got it, but if you want better visibility then, a dark background on the whole screen would be better I think. You can take inspiration from what the streaming apps do, you check various of them and apply whichever you find best.

Pentaphon:
> Okay got it, but if you want better visibility then, a dark background on the whole screen would be better I think. You can take inspiration from what the streaming apps do, you check various of them and apply whichever you find best.

This is probably the way to go. Its easy to resist this since the official app does this but it does work for pretty much all streaming services do this.

zoro11031:
Closing this PR to work on adding a dark gradient/screen dimmer with the playback overlay instead, per the discussion above. Thanks for the feedback.

damontecres:
I think putting a 50% semi-transparent black to 100% transparent gradient from bottom to top that covers ~1/3 of the screen would probably work.

The `controllerHeight` is calculated in `PlaybackOverlay`, so you could use that (maybe add a bit) to set the height of the gradient.

PR #496, opened 2025-12-17
Bump dependency versions, notably media3

### LINKED ISSUE #108: [BUG] - 4:3 and 2.35:1 videos sometimes have wrong scaling upon opening
### Description

This is a minor issue but I've noticed that 4.3 and 2.35:1 videos **sometimes** open with the wrong scaling, where the video displays outside the bounds of the TV, and then it corrects itself in a few seconds or when you bring up the Player UI. Both h264 and h265 show this error but never 16:9 videos, so I think it has to do with those aspect ratios. I think this is rarely noticed because most content is 16:9. 

This is never an issue with Jellyfin for Android TV so I don't think it is my device that is the issue because Jellyfin for Android TV also uses media3. However, it could be an issue with using an older version of media3? Not sure. 

I'm going to try to get video of this happening on my TV (or better yet, capture a log) but it happens enough that I think other users can reproduce this issue with the steps below. 

### Reproduction steps

1. Play a series of 4:3 and 2:35:1 videos: the more the better. 
2. Observe their scaling upon opening.
3. Eventually you will get one or more that opens with wrong scaling (the video displays outside the bounds of the screen) and it will either a) self-correct after a few seconds or b) self-correct when the player UI is displayed

### App Version

v0.2.2

### Server Version

10.11.1

### Device

Onn 4K Plus

### Logs

Added in https://github.com/damontecres/Wholphin/issues/108#issuecomment-3464655736

### LINKED ISSUE #355: [BUG] - Video Scale “Fit” is stretched
### Description

On all my testing I have found the default “Fit” video scale is stretched slightly and I have to change it to fit height or width depending on the source file aspect ratio. 

### Reproduction steps

_No response_

### App Version

V.0.3.3

### Server Version

10.10.7

### Device

Onn HD device 

### Logs

```bash

```

### LINKED ISSUE #479: [BUG] - Incorrect Aspect Ratio (2.35:1/Cinemascope) on playback start. Fixes when controls are displayed.
### Description

When playing any video with an ultra-wide aspect ratio (Cinemascope, 2.35:1 or 2.39:1, commonly labeled as 21:9), the image appears vertically stretched to fill the device's 16:9 format. This happens even when the display setting is set to "Fit" or "Full Width."

The problem is immediately resolved, and the image adjusts to the correct aspect ratio (e.g., 2.35:1 with black bars) as soon as the screen is tapped to **display the playback controls**. Once the controls fade away, the aspect ratio remains correct.

### Media info

Any video 21:9, HEVC, MKV

### Player backend

ExoPlayer (default)

### App Version

0.3.6

### Server Version

10.11.3

### Device

Google TV

### Logs

```bash

```

damontecres:
## Description
Bumps dependency versions, notably media3/ExoPlayer to `1.9.0`

### Related issues

Fixes #108
Fixes #355
Fixes #479


Pentaphon:
https://github.com/androidx/media/releases/tag/1.9.1 is out. Have you considered automating dependency updates or do you prefer to update manually?

Issue #436, opened 2025-12-12
[BUG] - Live TV EPG seems to be aligned on hourly boundaries


brianjmurrell:
### Description

It's now 9:45PM and I am looking at the live TV EPG.  It seems to be showing things that started at 9PM.onward.  There is a vertical line representing the current time running through the page, top to bottom.  But there are many shows that started at 9PM and were on for 30 minutes that are showing on the left of the guide.  But those shows have completely ended now and if I choose them of course I cannot choose to watch/record them.

Ultimately it seems pointless to have the EPG starting at 9PM when it's already 9:45PM and there are so many 30 minute programs that are already done and unwatchable.  It seems more reasonable to have the EPG starting at 9:30PM where the majority of programs are still viewable.  Or maybe it's most appropriate to have the guide starting at the current time with "already in progress" episode displays as described below being displayed.

Maybe this relates to #435 where some better way of displaying episodes that started in the past needs to happen.

I would propose that for episodes in progress (i.e. started before _now_) that the whole title be shown, perhaps prefixed with some kind of *back arrow* (i.e. `<` at minimum) to indicate that the episode started some time in the past.

### App Version

latest at time of writing

### Server Version

10.11.4

### Device

Chromecast sabrina

### Logs

```bash

```

Issue #435, opened 2025-12-12
[BUG] - Live TV EPG for episodes that started before "now"


brianjmurrell:
### Description

When scrolling through the EPG, if one scrolls onto an episode which started in the past, the title only shows (at best) the end of the title.  Usually this is unintelligible as to what the show title is as the beginning of it is truncated 

### App Version

latest at time of writing

### Server Version

10.11.4

### Device

Chromecast sabrina

### Logs

```bash

```

brianjmurrell:
#288 could go a ways to making this less horrible.  Ultimately it would be good to show the full title in the displayed portion of the time slot on the guide though.

Issue #434, opened 2025-12-12
[BUG] - TV shows in the EPG are too aggressively truncated


brianjmurrell:
### Description

While looking through the guide, a show that might have a 2 hour (so quite long, display-wise) time-slot has it's title ellipsized as if it was a 30 minute show with only the space show a 30-minute-slot title.

So for example for a 2h timelot you might see:

```
+----------------------------------------------------------------------------------+
|Foo Ba…                                                                           |
+----------------------------------------------------------------------------------+
```

where there is very clearly so much more unused white-space in which a more fulsome title could be shown.

### App Version

latest at the time of writing

### Server Version

10.11.4

### Device

Chromecast sabrina

### Logs

```bash

```

damontecres:
I don't see this behavior. Are you sure the guide data itself isn't truncated? How does it look on the web UI?

brianjmurrell:
Hrm.  Interesting.  The only occurrence I can find of it now is not so much a case of truncation, but a case of ellipisizing at the end of a title that fits.  So for example the box looks like:

```
+----------------------------------------------------------------------------------+
|Click…                                                                            |
+----------------------------------------------------------------------------------+
```

The XMLTV entry for that is:

```xml
        <programme start="20251212100000 -0800" stop="20251212123000 -0800" start_timestamp="1765562400" stop_timestamp="1765571400" channel="amc.us">
                <title>Click</title>
                <desc>An overworked architect meets an eccentric inventor who gives him a universal remote that allows him to control the world around him.</desc>
        </programme>
```

So as you can see, there is no need to ellipsize and doing so misleads one into thinking there should be more.

brianjmurrell:
Ah ha!  I found a bunch:

![Image](https://github.com/user-attachments/assets/6606aaa7-c1e6-4bd2-bef3-46a55766b0bf)

You can see 5 in that image starting with the highlighted entry.  The XMLTV for those are:

```xml
        <programme start="20251212110000 -0800" stop="20251212130000 -0800" start_timestamp="1765566000" stop_timestamp="1765573200" channel="[redacted]">
                <title>Pretty in Pink
 (1986)</title>
                <desc>An underprivileged Cinderella chooses between two prom dates: a charming preppy and a disarming buddy.. 
        ¤ Produced in: 1986. , 
        Movie, Comedy, Drama, Romance, Coming Of Age, Other. 
[...]

PR #521, opened 2025-12-20
TV Guide UI followup to #517

damontecres:
## Description
- Instead of just replacing newline characters, collapse white-space into single space characters which is roughly how the web UI does it with CSS's `white-space: nowrap;`
- Use an actual arrow indicator instead of a `<` character for when a program begins before the guide start

### Related issues
#517
https://github.com/damontecres/Wholphin/issues/434
https://github.com/damontecres/Wholphin/issues/435
https://github.com/damontecres/Wholphin/issues/436

### Screenshots
image


brianjmurrell:
Nice.

PR #398, opened 2025-12-08
Revert "Remove PIN & sign-in auto settings"

damontecres:
Reverts damontecres/Wholphin#397

PR #537, opened 2025-12-22
[FEAT] Add option to disable automatic sign-in on launch

zoro11031:
### Description

Previously, the app would always automatically sign in the last used user upon launch, preventing users from easily switching accounts without first fully loading the app, selecting their user in the top right,  and then switching.

This PR introduces a "Sign In Automatically" toggle in the Account settings to make this behavior optional.

**New Behavior:**

* **Toggle ON (Default):** Preserves the original behavior. The app automatically restores the session for the last used user.

* **Toggle OFF:** The app interrupts the auto-login flow to allow account selection.

* **Single Server:** If only one server is configured, the app skips the server list and navigates directly to the **User Selection** screen.

* **Multiple Servers:** The app defaults to the **Server Selection** screen.

* **Single User Exception:** If the target server has only **one registered user**, the toggle is ignored, and the app will auto-sign in to streamline access (as user selection is redundant in this case).

**Technical Implementation:**

* Added `SignInAuto` to `AppPreference`.

* Updated `MainActivity` to check preferences before session restoration.

* Added helper methods `hasMultipleUsersOnServer` and `getSingleServer` to `ServerRepository` to support the navigation logic.

* Updated `onRestart` to close the session if the user disabled auto-sign-in and multiple users exist.

### AI/LLM usage

* Used LLM to assist with the code logic for skipping the server selection screen when only a single server is present.
* Generated the PR description using an LLM based on `git diff` output.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
@@ -148,8 +148,7 @@ class MainActivity : AppCompatActivity() {
                             var singleServerForUserSelection by remember { mutableStateOf<JellyfinServer?>(null) }
                             LaunchedEffect(Unit) {
                                 val serverId = appPreferences.currentServerId?.toUUIDOrNull()
-                                val hasMultipleUsers = serverRepository.hasMultipleUsersOnServer(serverId)
-                                val shouldAutoSignIn = appPreferences.signInAutomatically || !hasMultipleUsers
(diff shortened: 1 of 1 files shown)


damontecres:
Thanks for the contribution @zoro11031!

Nearly all of this functionality is actually already implemented and just commented out. #398 has most of it.

I am planning to restore the auto-sign in setting, but separate from the PIN code for the next release.

PR #565, opened 2025-12-24
Allow debug builds to to be installed alongside the stable app

zoro11031:
## Description

This PR updates the debug build config to allow side by side installation with the mainline/stable release by changing the app name and app ID.

This is a convenience change to allow contributors to test debug builds on their personal devices without overwriting the existing stable version of the app

### Related issues

N/A

### AI/LLM usage

PR description drafted with AI assistance.

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -59,7 +59,6 @@ android {
             isMinifyEnabled = false
             isDebuggable = true
             applicationIdSuffix = ".debug"
-            versionNameSuffix = "-DEBUG"
         }
     }
     compileOptions {


damontecres (approved):
LGTM

Pentaphon:
@damontecres can we get explicit instructions somewhere like on the readme on how to install debug builds because right now the Downloader code is just trying to install itself over the Play Store version on my device.

zoro11031:
> @damontecres can we get explicit instructions somewhere like on the readme on how to install debug builds because right now the Downloader code is just trying to install itself over the Play Store version on my device.

Debug build would be different from building from  git/a bleeding edge release like you’re trying to get with the downloader code. You would manually build a debug apk using gradle on your system when you’re developing a feature or fix to test it. If you’re interested in doing that you can do it in Android Studio or via the command line by running ./gradlew assembleDebug in the project root

Pentaphon:
@zoro11031 only people with coding experience are going to do that. I was asking if we were going to get debug binaries for testers.

damontecres:
Yeah, this feature/PR is aimed towards developers. It makes it easier for devs to have (on the same device) both a stable build for normal use and also a dev build for testing their code.

It's unlikely that non-devs would want to use a debug build since they are slower and lag a lot. But the [develop branch builds](https://github.com/damontecres/Wholphin/releases/tag/develop) have debug APKs you could install manually if you really wanted to.

Pentaphon:
> It's unlikely that non-devs would want to use a debug build since they are slower and lag a lot. But the [develop branch builds](https://github.com/damontecres/Wholphin/releases/tag/develop) have debug APKs you could install manually if you really wanted to.

Thanks. I just spotted the debug builds before but should have updated with that. However, you may want to inform users on the main readme that they can now install a debug app alongside their Play Store install so you can hopefully get more people using your experimental branches.

Issue #562, opened 2025-12-24
[BUG] - <title> Slow Loading of Series Pages


thaneec:
### Description

I Discovered Slow Loading of Series Pages if i go into a Series, on Movies it's quit instant browse, but on Series especially with many Seasons Loading takes up to 10-15 seconds

### App Version

v.0.3.8

### Server Version

10.11.5

### Device

Android 11 Xiaomi TV-Stick 4k

### Logs

```bash

```

damontecres:
#567 greatly improves loading speeds for series.

But daily tv shows still need more work, so I will leave this issue open for now

thaneec:
Loads much faster now, if you go into the Season with lets say more as 10 episodes it takes long to load eg. 3-7 seconds
Thx for your awesome Work

damontecres:
There have been a couple PRs that improve this like #800

Also see #599

PR #567, opened 2025-12-24
Optimize series data loading speed

damontecres:
## Description
Optimizes how both series pages are loaded. If the series has a lot of seasons and/or lots of episodes per season, the pages load much faster.

This is a combination of reducing queried information, optimizing finding the right data to display first, and lazy loading seasons.

Note: Series that don't have episode numbers, like daily TV shows, are not as optimized yet

### Rough speed-ups
These numbers are from my (beefy) server, so it will vary based on your server

Series details page (one with seasons row) w/ ~50 seasons: 5s->300ms, ~13x faster

Series overview page (one with episode row) w/ ~50 seasons & 25 episodes: 6s->1.1s, ~5x faster

### Related issues
Related to #562


zoro11031:
Amazing job here, this was one of my biggest pain points and it’s pretty much completely fixed. Thank you!

Issue #570, opened 2025-12-25
[FEA] - smart subtitles not being auto selected


CraigWatt:
### Description

I have smart subtitles on for all my users on my jellyfin server, but I’ve noted that Wholphin does not adopt this default selection option, which goes a long way in selecting the “correct” subtitles for typical viewing experiences.

Is this missing?  It seems to work well on official jellyfin android tv client?

Is this a big or something yet to be implemented?

Any insight is appreciated.

May thanks,

Craig

CraigWatt:
![Image](https://github.com/user-attachments/assets/9895b8e7-f2d1-4cff-abc4-5ad073103788)

Note in JF client, where user has smart subtitles setting on for them, this would already have the default option pre selected.

The smart subtitles logic doesn’t seem to be trickling down to the pre selection of subtitle for users on Wholphin.

CraigWatt:
Difference between forced and default ^^ just FYI

zoro11031:
I think this could probably be fixed by having it do a check for the default flag and the word default when autoselecting a subtitle track if there’s 2 subtitle tracks for the same language. Ill take a look, will probably work it into the forced subtitle fix PR  #572  since they probably have similar root causes

CraigWatt:
Yeah, from a jellyfin server point of view the smart subtitles option for a given user is fantastic as 99% of the time, at least for me, it SELECTS the correct subtitle for typical viewing experience.

E.g an episode of a tv show where subtitles ARE expected for particular foreign dialog.

I don't know if smart subtitles also has logic built in server <_> client where it also determines if subtitles are TOO large therefore are abbots subtitles that are full subtitles rather than for typical viewing experience.

I basically need to find out exactly what the logic is, how that is typically applied to the clients decision logic etc etc.

I know for sure the jf server to jf android client works perfect in this regard.

Wholphin does everything else much better though.

I'm happy to do some pull request and try and have a stab at it after Christmas

J-o-h-n-M:
I tried googling but could not find it. What are "Smart subtitles"?  Is that a Wholphin thing or a Jellyfin thing?

CraigWatt:
![image](https://github.com/user-attachments/assets/34e99033-a644-4555-9c0c-e18e763d9680)

![image](https://github.com/user-attachments/assets/8531457f-0f2a-4ae5-b1fb-295a76f51f03)

J-o-h-n-M:
[...]

PR #572, opened 2025-12-25
[WIP] Fix auto select logic for forced subtitles

### LINKED ISSUE #571: [BUG] Forced subtitles not automatically selected
### Description:
I noticed that forced subtitles aren't being automatically selected during playback. This might be related to #570.

For example, on *PLUR1BUS* S01E06, there is an extended French sequence that requires forced English subtitles. These didn't turn on automatically, even though English audio was selected.

**Context:**
I suspect Wholphin's custom subtitle logic (which remembers settings per-show) is overriding the standard behavior. Even though Jellyfin's "Smart" subtitles usually handle this, Wholphin seems to prioritize my saved "Off" preference for the series over the file's forced flag.

**Expected Behavior:**
The client should automatically select a subtitle track if:

1. The track has the `forced` disposition flag.
2. (Fallback) The track name contains "forced" and matches the audio language.

**Logic Priority:**
This logic should take precedence over a saved per-series "None" preference, but it **should not** override other explicit subtitle preferences.

For example, in the *PLUR1BUS* scenario:

* If my preference is "None": **Select English Forced.**
* If my preference is "Spanish": **Stick with Spanish** (do not switch to English Forced).

### Reproduction steps

1. Open a series where your saved subtitle preference is currently "Off" or "None" (i.e., you haven't manually selected a subtitle track for this show previously).
2. Play an episode that contains a forced subtitle track (e.g., an English audio file with an "English Forced" subtitle track for foreign parts).
3. Observe the playback.
4. **Result:** The forced subtitles do not appear, and the subtitle selection remains "Off."
5. **Contrast:** If you play the same file in the standard Jellyfin web client (with "Smart" mode enabled), the forced track should be correctly selected.

### App Version

v0.3.8-4-g13370db

### Server Version

10.11.5

### Device

NVIDIA Shield TV Pro 2019

zoro11031:
## Description

This PR updates the subtitle selection logic in `StreamChoiceService` to correctly prioritize forced subtitle tracks, even when the user's series-level preference is set to "Off" or "None".

Previously, Wholphin's per-series subtitle memory would override the `forced` flag. If a user had subtitles disabled for a show, the player would enforce "Off" even for episodes containing forced subtitles (e.g., foreign language segments), requiring manual intervention.

**Changes Implemented:**

* Updated `chooseSubtitleStream` to check for forced tracks when `SubtitlePlaybackMode` is `NONE` or when the track index is explicitly `DISABLED` (e.g., from a saved resume state).
* Added `isForcedOrSigns` helper function to detect forced tracks via:
1. **Metadata:** The container's `forced` disposition flag.
2. **Name Fallback:** Checking if the track title contains "forced" or "signs" (case-insensitive), provided the track language matches the audio language.

* Added `findForcedTrack` helper to prioritize matching forced tracks to selected audio language.

Forced subtitles of the same language are displayed automatically without overriding explicit subtitle language choices (e.g., if a user explicitly selects Spanish subtitles, that preference is still respected over English forced tracks).

### Related issues

Fixes #571 

### Screenshots

N/A

### AI/LLM usage

LLM used to assist with drafting PR and write helpful comments/documentation

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackViewModel.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackViewModel.kt
@@ -730,11 +730,36 @@ class PlaybackViewModel
                         type = MediaStreamType.SUBTITLE,
                     )
                 this@PlaybackViewModel.currentItemPlayback.setValueOnMain(itemPlayback)
+
+                // Resolve ONLY_FORCED to an actual track index for immediate playback
+                // (The preference is already saved as ONLY_FORCED for future resume)
+                val resolvedIndex =
+                    if (index == TrackIndex.ONLY_FORCED) {
+                        val source = currentPlayback.value?.mediaSourceInfo
+                        val audioStream =
+                            currentMediaInfo.value?.audioStreams?.firstOrNull {
+                                it.index == itemPlayback.audioIndex
+                            }
+                        val candidates =
+                            source?.mediaStreams
+                                ?.filter { it.type == MediaStreamType.SUBTITLE }
+                                .orEmpty()
+                        streamChoiceService.chooseSubtitleStream(
+                            audioStream = candidates.firstOrNull { it.index == audioStream?.index },
+                            candidates = candidates,
+                            itemPlayback = itemPlayback,
+                            playbackLanguageChoice = null,
+                            prefs = preferences,
+                        )?.index
+                    } else {
+                        index
+                    }
+
                 changeStreams(
                     item,
                     itemPlayback,
                     itemPlayback.audioIndex,
-                    index,
+                    resolvedIndex,
                     onMain { player.currentPosition },
                     true,
                 )


zoro11031:
Might be worthwhile to add a toggle on/off to the settings menu for this behavior since it overrides your sub settings if you have none selected.

zoro11031:
Added a quick fix for #570 by ensuring that default tracks are always prioritized over others when matching a subtitle language

damontecres:
Hey, thanks @zoro11031. I think I want to go a slightly different direction. I think separate fixes for #570 & #571 would be good.

I'm working on fix to address #570 which basically updates the smart stream selection logic to more closely follow the server's.

I think for #571, adding an option for "Only forced" when selecting subtitles might be an decent option and add it as another special case to `TrackIndex` object. I don't think a global setting should be needed.

Separately though, I am wondering though if trying to have Wholphin assume/guess what the series should use for subtitles is ultimately not best user experience. It is opaque and not super obvious to new users. I wonder if instead having an option to pick the subtitle mode & language per series makes more sense. I think I'll file an issue to discuss this idea more.

zoro11031:
> Hey, thanks @zoro11031. I think I want to go a slightly different direction. I think separate fixes for #570 & #571 would be good.
> 
> I'm working on fix to address #570 which basically updates the smart stream selection logic to more closely follow the server's.
> 
> I think for #571, adding an option for "Only forced" when selecting subtitles might be an decent option and add it as another special case to `TrackIndex` object. I don't think a global setting should be needed.
> 
> Separately though, I am wondering though if trying to have Wholphin assume/guess what the series should use for subtitles is ultimately not best user experience. It is opaque and not super obvious to new users. I wonder if instead having an option to pick the subtitle mode & language per series makes more sense. I think I'll file an issue to discuss this idea more.

That makes sense to me! Did you want me to rework this PR and narrow the scope back to the forced subtitle behavior? Or would you rather me close it pending discussion?

My thoughts on having it be a global setting were that often times when watching a movie or show, you don’t know you need forced subtitles turned on until you get to a foreign language scene. And sometimes shows have foreign languages being spoken without subtitles at all as a stylistic choice. So for me personally, I think it’s disruptive to have to check per movie and per show (and possibly per episode if not every episode has forced sub scenes), because often times you’d find yourself pausing playback to check for a forced subtitle file.

I suspect for 90% of use cases you would want forced subtitles turned on any time they’re available as well, and if you want to turn them off because you coincidentally know the language being spoken, that can be a per show override handled by the current logic. So maybe just have it silently have them on by default when available per show, but it remembers if you turned them off for a specific show? The “only forced” option you proposed sounds like it would be a really elegant solution for this. 

I also think that for anime dub watchers, the logic I added to recognize and play Signs & Subs tracks would be a big quality of life improvement. Those tracks are often not flagged as forced even when they should be.

I hope I’m making sense. Thanks for all the work you put in to Wholphin!

zoro11031:
I see your recent merge that adds some more logic for forced subtitles and other subtitle scenarios - gonna just close this PR and work on adding the only forced subtitles option :)

Issue #552, opened 2025-12-23
[BUG] - <title>Paused Video stops when sceensaver pops up.


unimatrix72:
### Description

When i pause a video and then leave the room for 5 mins, after that i can't continue playing with pressing pause again.
It is stopped, and the video has to be selected and pressed again to continue.

### Reproduction steps

1. Set your Screensaver to lowest possible value (mine is on 5 mins. It's Areal Dream)
2. Start a video
3. Pause it
4. Wait until the screensaver pops up.
5. Press pause to countinue.
6. The "cursor" is left in the menu where home and Favorites etc. are. Then the video has to be selected again to continue playing.

### App Version

v0.3.8-0-g86e5e5a

### Server Version

10.10.7

### Device

SkyworthDigital SKWHP44H-ZWL Android TV 14

### Logs

```bash
Send App Logs
clientInfo=ClientInfo(name=Wholphin, version=0.3.8-0-g86e5e5a)
deviceInfo=DeviceInfo(id=c91563b7d240653e, name=Android TV)
--------- beginning of main
12-23 21:24:46.594  2852  2915 V Wholphin: saveActivity: itemId=9188b485-093c-227e-a19b-0d2d2dfabece, pos=1882396
12-23 21:24:46.595  2852  2915 I OkHttpClient: POST http://192.168.2.9:50000/Sessions/Playing/Progress
12-23 21:24:51.593  2852  2912 V Wholphin: saveActivity: itemId=9188b485-093c-227e-a19b-0d2d2dfabece, pos=1882396
12-23 21:24:51.596  2852  2912 I OkHttpClient: POST http://192.168.2.9:50000/Sessions/Playing/Progress
12-23 21:24:56.594  2852  2912 V Wholphin: saveActivity: itemId=9188b485-093c-227e-a19b-0d2d2dfabece, pos=1882396
12-23 21:24:56.596  2852  2912 I OkHttpClient: POST http://192.168.2.9:50000/Sessions/Playing/Progress
12-23 21:25:01.594  2852  2912 V Wholphin: saveActivity: itemId=9188b485-093c-227e-a19b-0d2d2dfabece, pos=1882396
12-23 21:25:01.596  2852  2912 I OkHttpClient: POST http://192.168.2.9:50000/Sessions/Playing/Progress
12-23 21:25:06.594  2852  2915 V Wholphin: saveActivity: itemId=9188b485-093c-227e-a19b-0d2d2dfabece, pos=1882396
[...]

PR #577, opened 2025-12-26
Further improvements to resuming the app

damontecres:
## Description
Fixes issues when the app is resumed, such as after the screensaver is dismissed.

For example, if you pause during playback and the screensaver starts, Wholphin will stop playback and go back to the previous page. But with this PR, instead of focus being lost and defaulting to teh nav drawer, focus will restore on the play button or episode card.

### Related issues
Related to #552 & https://github.com/damontecres/Wholphin/issues/552#issuecomment-3690614491


Hukuma1:
The focus goes to the nav bar also when a movie ends I believe. Could that also please be tweaked? (If not already part of this PR.)

damontecres:
> The focus goes to the nav bar also when a movie ends I believe. Could that also please be tweaked? (If not already part of this PR.)

Yes, this PR handles that as well.

PR #586, opened 2025-12-27
Minor UI tweaks

### LINKED ISSUE #431: [FEA] - Setting to change what's shown for Recently Added Series episode counts
### Description

On any library set to series, the recently added displays shows with their total episode count. I think this should show the user's unwatched count for each series instead. 

I wasn't sure if this should be a bug to fix this, or a feature request to add a setting to switch between the two options.

damontecres:
## Description
- Show unwatched counts on the cards in the latest rows instead of child count, which is more in line with the web UI & official app
- Show exact runtime on the overview dialog for each version
- Don't show controls after clicking up next

### Related issues
Closes #431

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/data/ItemDetailsDialogInfo.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/data/ItemDetailsDialogInfo.kt
@@ -32,6 +32,7 @@ import org.jellyfin.sdk.model.api.MediaStream
 import org.jellyfin.sdk.model.api.MediaStreamType
 import org.jellyfin.sdk.model.api.VideoRange
 import org.jellyfin.sdk.model.api.VideoRangeType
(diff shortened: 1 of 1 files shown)


lbux:
I'm not sure if it was this commit or another one but when a show is completed and removed from next up, cornerText seems to not update consistently with the cards. cornerText seems to reference the previous structure of the list of cards. For example, if card 1 was at E10 of Show 1 when it home page was last accessed and you finish all remaining episodes and go back to the home page, card 1 now is now Show 2 but has the cornerText of Show 1 (E10). I've only been able to recreate this once, but if it happens again I will make a proper bug report.

damontecres:
> not sure if it was this commit or another one but when a show is completed and removed from next up, cornerText seems to not update consistently with the cards

Yes, I see this bug and this PR certainly caused it. I'll push a fix later today.

PR #589, opened 2025-12-28
Add "Only Forced Subtitles" option to subtitle selection

### LINKED ISSUE #571: [BUG] Forced subtitles not automatically selected

zoro11031:
## Description

  Adds a client-side "Only Forced Subtitles" option when selecting subtitles, per the discussion in #571.

  ### Changes

  - **TrackIndex.ONLY_FORCED** (`-3`): New constant for the special case
  - **StreamChoiceService**: Intercepts `ONLY_FORCED` at the top of `chooseSubtitleStream()` before server-side `SubtitlePlaybackMode` logic (which remains untouched)
  - **UI**: Adds "Only Forced Subtitles" option to both:
    - In-player subtitle menu (`PlaybackDialog.kt`)
    - Pre-playback stream selection dialog (`Dialogs.kt`)

  ### Forced track detection

  Uses metadata `isForced` flag as primary check, with title-based fallback ("forced", "signs", "songs" patterns) when language matches audio.

  ### Related issues

  Closes #571

  ### Screenshots

  N/A - subtitle selection menu only

  ### AI/LLM usage

  Used Claude Code for assistance and to draft PR description.


app/src/main/java/com/github/damontecres/wholphin/services/StreamChoiceService.kt
```diff
@@ -259,6 +262,45 @@ class StreamChoiceService
...
+
+            return false
+        }
+
+        /**
+         * Finds a forced/signs track, preferring those matching the audio language.
+         */
+        private fun findForcedTrack(
+            candidates: List<MediaStream>,
+            audioLanguage: String?,
```
damontecres: I'm wondering if this should be the user's preferred subtitle language, maybe use stream audio for sign-subtitles if they have no preference.

Some users in other issues have mentioned preferring different audio and subtitle languages.

Also maybe need to account for subtitles with unknown subtitles and fallback to those as well. You can use `MediaStream.language.isUnknown` for check that.

damontecres (commented):
I pushed a couple fixes for things. I had some issues still, but it seemed better once I cleared the track selections.

So I still want to test this more, but I think it's almost there.

zoro11031:
> I pushed a couple fixes for things. I had some issues still, but it seemed better once I cleared the track selections.
> 
> So I still want to test this more, but I think it's almost there.

Looks good! I should have some time to test it tonight on my end too.

zoro11031:
Made a small tweak that I think should hopefully help fix some of the finnickiness. Need to test more to be sure

damontecres (approved):

damontecres:
I merged in the changes from #664 and I think this PR is ready!

PR #618, opened 2026-01-02
Show button for trailers instead of card row

### LINKED ISSUE #408: [FEA] - Access movie trailers more easily
### Description

Unless I'm not seeing it, could you please add the option to watch movie trailers when a movie is selected?

damontecres:
## Description
On both movie & series detail pages, show a button to play trailers instead of the row of cards below.

The button works like this:
- If no trailers, the button shows "No trailers" and is disabled/non-clickable
- If only a single trailer, the button shows "Play trailer" and clicking it plays the trailer
- Otherwise, shows "Trailers" and clicking displays a dialog to choose one

Local trailers are shown first in the list. Then there is an attempt to sort the trailers for the dialog so that names with "Official Trailer" or "Official Theatrical Trailer" are listed first and "Teaser" is listed last. This isn't perfect and doesn't work for non-English, but there's no other metadata to use.

### Related issues
Closes #408

## Screenshots
trailer button
trailer list

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/components/Dialogs.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/components/Dialogs.kt
@@ -9,6 +9,7 @@ import androidx.compose.foundation.gestures.scrollBy
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.BoxScope
+import androidx.compose.foundation.layout.Column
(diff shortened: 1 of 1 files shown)


Pentaphon:
Looks great and you thought of everything I would have suggested. It's amazing how you can make something happen within days that the official app takes months to do or never does after a feature request!

The only design issue I see is that the selected trailer is a little too close to the "play trailer" window title so a little more padding between the list and the title would make it look perfect. Thanks!

PR #644, opened 2026-01-07
Fix suggestions mixing content from different libraries and update recommendation logic

### LINKED ISSUE #638: [BUG] Suggestions in library view mix all show libraries together
### Description

When a server is configured with multiple libraries of the "Shows" type (e.g., a "TV Shows" library and an "Anime" library), the "Suggestions" view fails to limit results to the currently selected library. Instead, it pulls content from all libraries of that type.

Screenshots:

Screenshot 1: Shows the "Anime" library is currently selected.

Image

Screenshot 2: Shows content from the "TV Shows" library appearing in the "Anime" suggestions view.

Image

_(Note: I have confirmed that the non-Anime shows are NOT present in the source folders mapped to the Anime library.)_

### Reproduction steps

1. Create two distinct libraries with the content type set to "Shows" (e.g., Library A and Library B).

2. Ensure source folders for both libraries are separate and contain unique content.

3. Navigate to Library A.

4. Scroll to or select the "Suggestions" view.

### App Version

v0.3.11

### Server Version

10.11.5

### Device

NVIDIA Shield TV Pro 2019

### Logs

```bash

```

zoro11031:
## Description
This PR aims to fix #638 where the "Suggestions" row was mixing up content from different libraries that shared the same library type. For example you were in the Anime library, you'd see regular TV shows showing up, as the TV Shows and Anime libraries would both be "Show" libraries.

As @damontecres mentioned in the issue discussion, the `/Items/Suggestions` endpoint Wholphin is using just returns random items from the same library type, with no special logic or recommendation algorithm.

I also looked at how the official WebUI handles this. It basically just mixes "Resume," "Next Up," and "Latest Items" together. That felt a bit redundant since we usually have dedicated rows for those things anyway; I wanted to try something that helps discover new content.

I replaced the broken endpoint with a few custom `GetItemsRequest` calls. These work correctly with the `ParentId` filter, so we only get results from the library we are actually looking at. Additionally, I added in some logic that takes into account what the user has been watching to tailor the suggestions a little bit.

The new logic uses a 40/30/30 mix (this can of course be tweaked as needed)

- Tailored (40%): Grabs your recently watched items from this library, checks their genres, and returns items from the same genre.

- Random (30%): Random unwatched stuff from this library.

- New (30%): Recently added items.

This is implemented using only `GetItemsRequest` API calls using filters, rather than any specialized endpoints like the `/Items/Suggestions` endpoint. This means the feature should be predictable and stable. If Jellyfin ever adds an actual recommendation algorithm and endpoint to call it, I imagine we would want to switch to that in the future though.

I also added some failsafe logic to make sure there will be diversity in the "recently watched" items that get pulled, in case a user might have recently binged a TV show. We wouldn't want every single recommendation to be based off of that one show. 

To fix this, I changed the query to fetch the last 20 history items instead of just the top 3. Then, we filter that list client-side to find unique Series IDs. If the user watched 10 episodes of One Piece in a row, the logic collapses them into a single "One Piece" entry and moves on to the next distinct show they watched. This guarantees we get variety in the source material for recommendations.

This is a work in progress right now, mainly due to performance. For my server it takes about 10-15 seconds to load the Suggestions row when I tested. This is unacceptable in my opinion, especially if we ever intend to add a Suggestions row like this to the home page like Plex does. I have some ideas on how to improve performance, but I'm open to suggestions.

### Related issues
Resolves #638 

### Screenshots

Before:

suggestions1

After:

suggestions3

### AI/LLM usage
I used Claude Code to help write the Kotlin Coroutines so the API calls run at the same time. I also used it to write the tests in `TestSuggestionsLogic.kt`.


damontecres (commented):
I haven't tested or anything, but at a glance, looks like a lot of duplicated code between the two view models with just different types. Maybe move the logic to a service that takes a `BaseItemKind` parameter?

I do know there's already quite a bit of duplicated code between them, but I just want to keep that from ballooning even more if we can.

zoro11031:
> I haven't tested or anything, but at a glance, looks like a lot of duplicated code between the two view models with just different types. Maybe move the logic to a service that takes a `BaseItemKind` parameter?
> 
> I do know there's already quite a bit of duplicated code between them, but I just want to keep that from ballooning even more if we can.

Yeah probably a good idea. I'll move the suggestions logic to its own service.

zoro11031:
I think I've been able to address the worst of the performance bottlenecks. The biggest change was refactoring the ViewModel to fetch all the home screen rows in parallel instead of waiting for them to load one by one. That alone should make a huge difference in how fast the screen fills up.

I also added a small cache service that caches the suggestions from the last session, so that the Suggestions row can populate immediately without waiting for fresh data. It just refreshes in the background now, which feels like a safe bet since suggestions shouldn't change drastically day to day.

I’ll push the changes later today, I want to test it a bit more first. The server row looks like it's populating correctly (it's at least not mixing libraries), but I want to check the server side debug logs just to confirm the API is actually getting queried correctly and serving everything we'd expect it to.

zoro11031:
I added memory caching to the cache component. Not super used to working on stuff like that, so I did have to lean on AI a lot to double check it (very paranoid about causing a memory leak or destroying performance on low end devices). I can remove the memory caching if it seems like overkill/unnecessary.

zoro11031:
I need to test more on a physical device and review the logs still, but from my testing in Android Studio I think it's shaping up pretty good

app/src/main/java/com/github/damontecres/wholphin/services/SuggestionsCache.kt
```diff
@@ -0,0 +1,107 @@
...
+            .resolve("${cacheKey(libraryId, itemKind)}.json")
+
+        suspend fun get(
+            libraryId: UUID,
+            itemKind: BaseItemKind,
+        ): CachedSuggestions? {
+            val key = cacheKey(libraryId, itemKind)
+            memoryCache[key]?.let { return it }
+
+            return withContext(Dispatchers.IO) {
```
damontecres: This will read the cache file on every cache miss. Probably just want to read it once either on the first miss or on start up.

Maybe add to `PlaybackLifecycleObserver` (not a great name) `onCreate` to initialize the cache.

damontecres (commented):
This is an interesting idea for the implementation.

I wonder if trying to work this into the regular page loading workflow is not the best approach though. What do you think of having a background worker periodically fetching suggestions and caching those instead? It could run once or twice a day maybe.

`TvProviderWorker` works this way to add the items to the OS "Watch Next" row.

zoro11031:
> This is an interesting idea for the implementation.
> 
> I wonder if trying to work this into the regular page loading workflow is not the best approach though. What do you think of having a background worker periodically fetching suggestions and caching those instead? It could run once or twice a day maybe.
> 
> `TvProviderWorker` works this way to add the items to the OS "Watch Next" row.

I was mulling over a similar idea too. It would prevent the possible issue of a user browsing the cached data on start up, only for it to update while they're looking. I wouldn't imagine that updating suggestions only twice a day or even once a day would be an issue for anyone realistically. Should also make everything snappier now that it's not running the suggestions logic every start up. I'll look into that and also address your feedback this weekend.

zoro11031:
@damontecres I've addressed all your comments in the latest 3 commits. I'm working on the idea to separate out the syncing/caching to a background worker on a separate branch that I'll merge into this one later once I have it working :).

zoro11031:
@damontecres I just merged in the background worker branch to this PR :)

Quick rundown: suggestions now get fetched via a background worker (every 12h, or immediately if cache is empty). The worker fetches all libraries in parallel using supervisorScope/async, which cuts down wake time significantly. It should help a lot on lower end hardware like Fire TV sticks.

Also refactored SuggestionsCache to use Mutex with suspend functions so we're not doing any disk I/O on the main thread. Added exponential backoff for retries so we're not hammering the server if something goes down.
Unit tests are there for the worker, scheduler, and cache concurrency stuff.

zoro11031:
Hmm, those unit tests that failed in the CI weren’t failing when I ran debug build on my end. I’ll look into that today.

zoro11031:
Yeah the unit tests were not very good - relied too much on AI to write them ("they're just tests, what could go wrong, etc"). The AI overcomplicated them and made them unreliable. I massively trimmed them down. I'd rather have a little bit of reliable test coverage than a massive amount of garbage

damontecres:
Hey @zoro11031 is this ready for a review?

zoro11031:
> Hey @zoro11031 is this ready for a review?

Yep, all good! Forgot to run the linter so I just ran that. Pre commit CI should pass now

zoro11031:
I'll look into the compilation error later, it was compiling for me so it might be a typo or stupid mistake on my end. Sorry for being a bit less responsive this week, been a busy week with the family

zoro11031:
@damontecres should build now! It was failing because I accidentally deleted imports for the seerr integration when merging main back into the branch. doy

damontecres:
Looks pretty good. I made a few changes, so please give them a quick look. Most were just code style and using stream to parse the JSON.

But I did update how `SuggestionCache.loadFromDisk` works. It now takes a userId and gets the files for that user. And if a new userId is requested, it clears the memory cache and reloads that user. Should work the same, but use less memory.

damontecres (approved):

zoro11031:
Lgtm. Thanks for the assist on the cache logic.

Side note, any thoughts on adding “Suggestions” as one of the home row options after #803 is merged?

damontecres:
> Side note, any thoughts on adding “Suggestions” as one of the home row options after https://github.com/damontecres/Wholphin/pull/803 is merged?

Yup, I was planning to add it once this PR is merged!

zoro11031:
@damontecres I'm seeing performance regressions post merge that I wasn't seeing when testing. In particular start up time is severely impacted. I'm working on diagnosing it now. The suggestions fetch logic shouldn't be blocking so not sure why loading the entire homepage is being impacted

PR #652, opened 2026-01-07
Show indicator for multiple versions of media file

### LINKED ISSUE #612: [FEA] - Version Handling Improvements
### Description

First of all, let me say how much I’m grooving on Wholphin. I’m using it more often than any other client. However, I have a feature request.

I’d like to see better handling of versions. With the current version of Wholphin, I have to know a movie has different versions, there is no indication when scrolling through the library. And if I remember there are different versions I then have to navigate, on the movie page, to the ellipses, choose versions, and pick the version I want to watch.

Please consider a visual indication on the movie poster that there are multiple versions of a movie, and list/make selectable from the main movie info page once a user has clicked into it from the library.

I initially posted this in discussions, not realizing this is the forum for feature requests. Sorry for the duplicate.

damontecres:
## Description
Adds an indicator to cards on grid pages if there are more than a single version of the item. Also adds the count to the video stream label, much like for multiple audio/subtitle tracks.

No behavior changes, just visual changes. You still need to select the version from More->Choose Version.

### Related issues
Closes #612

### Screenshots
image
image

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/components/VideoStreamDetails.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/components/VideoStreamDetails.kt
@@ -47,7 +47,7 @@ import org.jellyfin.sdk.model.api.VideoRangeType
 @NonRestartableComposable
 fun VideoStreamDetails(
     chosenStreams: ChosenStreams?,
-    numberOfVersions: Int = 0,
+    numberOfVersions: Int,
     modifier: Modifier = Modifier,
 ) = VideoStreamDetails(
     chosenStreams?.source,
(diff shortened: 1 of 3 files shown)


datwildfire:
I really love this feature!

Is there currently any way to tell the client to **always play the highest available quality** for a file (for example prefer 4K when it exists), or is that something that would need a separate option?

Basically looking for a set-and-forget way to say: *“always pick the best stream this file has.”*

damontecres:
> Is there currently any way to tell the client to **always play the highest available quality** for a file (for example prefer 4K when it exists), or is that something that would need a separate option?
> 
> Basically looking for a set-and-forget way to say: _“always pick the best stream this file has.”_

I replied to you in https://github.com/damontecres/Wholphin/issues/700#issuecomment-3751649399

PR #670, opened 2026-01-11
Add option to always direct play Dolby Vision Profile 7

### LINKED ISSUE #449: [BUG] - Dolby Vision Profile 7 gets remuxed to HDR10
### Description

The Nvidia Shield Pro 2019 is capable of playing back Dolby Vision profile 7. If the video includes the enhancement layer (EL), the Nvidia Shield will correctly handle it internally by playing the DV part without the EL. This is what happens with the old Jellyfin 10.10 version, or with other video players (Kodi, Just Player etc...)

However, apparently the device is reporting that it doesn't support EL, and Jellyfin 10.11 is then remuxing the file to completely remove the DV layer:
```
<summary>Codec HDR Support</summary>

**HEVC**
  - Dolby Vision: true
  - Dolby Vision Enhancement Layer: false
  - HDR10: true
  - HDR10+: false

[...]

### Device information

***Android version***: `11`  
***Device brand***: `NVIDIA`  
***Device product***: `mdarcy`  
***Device model***: `SHIELD Android TV`  
***Device manufacturer***: `NVIDIA`  
***Device codename***: `mdarcy`  
```

This issue has been reported in https://github.com/jellyfin/jellyfin-androidtv/issues/5091, with a proposed solution https://github.com/jellyfin/jellyfin-androidtv/pull/5116.

It would be good to do something similar in Wholphin as well.

Probably one option is to add the "Force Dolby Vision Profile 7" under the advanced menu. While another option could be to add a special case for the Nvidia Shield (`mdarcy`) to always set Dolby Vision EL as supported (a kind of quirk similar to the `KnownDefects.kt` list).

### Reproduction steps

1. Try to playback a video that has DV 7 with EL
2. Notice that Jellyfin will remux the file and completely remove the DV layer

### App Version

v0.3.5

### Server Version

10.11.4

### Device

NVIDIA Shield TV Pro 2019

### Logs

```bash
{
            "Type": "Video",
            "Conditions": [
                {
                    "Condition": "NotEquals",
                    "Property": "VideoRangeType",
                    "Value": "DOVIInvalid|DOVIWithEL|DOVIWithELHDR10Plus|HDR10Plus",
                    "IsRequired": false
                }
            ],
            "ApplyConditions": [
                {
                    "Condition": "EqualsAny",
                    "Property": "VideoRangeType",
                    "Value": "DOVIInvalid|DOVIWithEL|DOVIWithELHDR10Plus|HDR10Plus",
                    "IsRequired": false
                }
            ],
            "Codec": "hevc"
        },
```

RyuzakiKK:
## Description
A few devices, like the Nvidia Shield Pro 2019, are able to correctly handle Dolby Vision Profile 7. However, when a Full Enhancement Layer (FEL) is present, these devices effectively ignore the enhancement layer and only apply the RPU metadata.

Due to this behavior, device capabilities are reported as not supporting `DOVIWithEL`, which causes the Jellyfin server to strip the DV layer entirely with a remux.

This commit adds an option to force direct play of Dolby Vision Profile 7 content, preventing the server from stripping the Dolby Vision layer and allowing compatible devices to handle playback as intended.

### Related issues
Fixes https://github.com/damontecres/Wholphin/issues/449

### Screenshots
dovi7

### AI/LLM usage
None


damontecres (approved):
I'd like to eventually figure out a nice way for configuring direct play overrides that is not just a huge list of toggles.

But for now, this is good. Thanks for the contribution!

Phil57:
I was battling for an hour with transcoding without reason and this fixed it !
Won’t watch my movie tonight but it will be fine next time.

Thanks !

PR #678, opened 2026-01-12
Feat(ui): Add drop shadow to UI dialogs

### LINKED ISSUE #665: [UI] Better dialog effect
Side note, I think the dialogs on regular pages would look better with a drop shadow or some effect to give "elevation". They look awkward and flat.

I'll open an issue for this since I don't want to do it in this PR.

_Originally posted by @damontecres in https://github.com/damontecres/Wholphin/issues/664#issuecomment-3731398730_

zoro11031:
## Description
Adds a subtle drop shadow to dialog containers in Dialogs.kt. Should make the dialogs pop a bit more instead of looking flat.

### Related issues
Fixes #665

### Screenshots

Before:

before1

After:
after

### AI/LLM usage
None


Pentaphon:
Wow, I should have thought to request this. Nice fix. It's a small detail that does make a difference.

damontecres (approved):

PR #688, opened 2026-01-13
Fix wrong dialog showing after Seerr login

zoro11031:
**Description**

Fixes a race condition in the Seerr integration.

When you added a new Seerr server and logged in successfully, the app would immediately show the "Disable Seerr?" dialog instead of a success message. The state was getting mixed up during recomposition and thinking you wanted to remove the server you just added.

Fixed by being more explicit about what mode the dialog is in (adding vs removing) so it can't get confused by background state changes. Also added a success toast when login works.

**AI/LLM usage**

Used Claude Code to analyze the issue.


damontecres:
This was on my list, great job!

damontecres (approved):

PR #709, opened 2026-01-16
Search page improvements

zoro11031:
## Description

Improves how search results are displayed:

  - Reorders search categories to Movies → TV Shows → Episodes → Collections, surfacing the most commonly searched content types at the top.   
- Adds a new "Combined search results" preference in Advanced settings → Search. When enabled, shows all results in a single list instead of separate categories. 
 - When Seerr is active, adds Library/Discover tabs to switch between local library results and Seerr discovery results.
 -   Hides empty result categories instead of showing "No results" placeholders.

### Screenshots

Screenshot_20260116_113656

image

image

### AI/LLM usage
Code auditing and PR description drafting with Claude Code. Also used Claude Code for the regex.


damontecres:
Just curious on the use case for combining results. Is it just a personal preference?

The reordering is a good idea. I don't know why I put collections second.

zoro11031:
> Just curious on the use case for combining results. Is it just a personal preference?

Pretty much. I've noticed that some streaming services split results into categories, and some mix them all together. Plex mixes them all together by default but has a tab to switch to the category view.

For me personally I prefer combined because when I'm searching I generally already know what I'm looking for, my search is going to closely match the title, so I prefer it to be the first result regardless of what category it falls under, if that makes sense.

damontecres:
That makes sense. What do you think about adding a "View Options" button (like on library pages) on the search page and having the toggle there instead? Maybe to the right of the search text box? I'm trying to keep the settings pages from being an endless list of toggles.

For the combined results, I think putting them in a `CardGrid` would be better than a single row.

app/src/main/java/com/github/damontecres/wholphin/ui/main/SearchPage.kt
```diff
@@ -158,6 +176,41 @@ class SearchViewModel
...
+                            fields = SlimItemFields,
+                            limit = 50,
+                        )
+
+                    val result = api.itemsApi.getItems(request).content
+                    val items =
+                        (result.items ?: emptyList()).map {
+                            BaseItem.from(it, api, false)
+                        }
+                    val sorted = items.sortedWith(compareBy { SearchRelevance.score(it, query) })
```
damontecres: Would be cool to apply this to the uncombined results too

zoro11031:
> That makes sense. What do you think about adding a "View Options" button (like on library pages) on the search page and having the toggle there instead? Maybe to the right of the search text box? I'm trying to keep the settings pages from being an endless list of toggles.
> 
> For the combined results, I think putting them in a `CardGrid` would be better than a single row.

Yeah, both of those thoughts make sense to me. I remember you're pretty wary of adding more toggles to the settings menu, haha.

app/src/main/java/com/github/damontecres/wholphin/ui/main/SearchPage.kt
zoro11031: Good idea! I'll get on that

app/src/main/java/com/github/damontecres/wholphin/ui/main/SearchPage.kt
```diff
@@ -158,6 +176,41 @@ class SearchViewModel
...
+        private fun searchCombined(query: String) {
+            viewModelScope.launch(ExceptionHandler() + Dispatchers.IO) {
+                try {
+                    val request =
+                        GetItemsRequest(
+                            searchTerm = query,
+                            recursive = true,
+                            includeItemTypes =
+                                listOf(
+                                    BaseItemKind.MOVIE,
```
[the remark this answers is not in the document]
damontecres: That's fair. I feel like it would be uncommon for someone to search for an episode specifically, at least I never have.

But including collections would be good.

zoro11031:
@damontecres Just finished up with the changes based on your review. I moved the toggle from the settings page to the search page, applied the relevance scoring to both search options, and switched the combined results to show using cardgrid. 

image


image

image

image

zoro11031:
@damontecres Coming back to this one now that the busy season at work is over, is there anything you need on my end to get this merged? Let me know! :)

damontecres:
@zoro11031 Sorry for delays on this. Definitely want to come back to this because the search page needs improvement and this is probably 90% of the work done.

I'll have a look again this week or next week.

damontecres:
@zoro11031 I made some UI adjustments and switch discover to show as a grid if combined is enabled.

Have a look! I think this is ready to merge. I have some other changes I want to do, but I will do them in a follow up PR.

damontecres (approved):

zoro11031:
> @zoro11031 I made some UI adjustments and switch discover to show as a grid if combined is enabled.
> 
> Have a look! I think this is ready to merge. I have some other changes I want to do, but I will do them in a follow up PR.

Lgtm. Thanks for the help on it :)

PR #734, opened 2026-01-21
Fix Seerr Local/Jellyfin authentication and request permissions

voc0der:
## Description
When using LOCAL or JELLYFIN authentication methods to connect to Seerr, the login flow was not properly establishing session cookies and user permissions. The login would 'succeed' and you were led to thinking it worked, but then...

This caused:
- Request functionality to be greyed out even though the user should have access
- Users to appear to have no permissions after login (permissions=0)

~~Additionally, during connection testing, passwords were being sent as API key headers regardless of the auth method.~~

## Solution
This PR fixes the authentication flow by:

~~1. **Only sending X-Api-Key header for API_KEY auth**: Modified `testConnection()` to only pass the credential as an API key when `authMethod == SeerrAuthMethod.API_KEY`. This prevents passwords from being accidentally treated as API keys during connection tests.~~

2. **Fetching full user config after login**: For LOCAL and JELLYFIN auth methods, the login endpoints (`authLocalPost()` and `authJellyfinPost()`) establish the session cookie but return incomplete user data. This PR now calls `authMeGet()` immediately after login to fetch the fully-populated user configuration including proper permissions.

## Tests
- Have working apk on my dev fork.
- Ran logins as Local + API. Cannot test Jellyfin due to environment, but codewise, it looks correct.

## Related issues
Related to https://github.com/damontecres/Wholphin/issues/747

## AI/LLM usage
Claude + ChatGPT usage


damontecres:
Thanks for the contribution!

I do see how the test connection incorrectly uses the password as the API key, so thank you for the fix for that.

Can you elaborate on the "Users to appear to have no permissions after login (permissions=0)" issue? Such as what version of Jellyseerr or any other information. I'm not enable to reproduce this behavior.

voc0der:
I will build it locally without the fix soon to verify, but I am using the `docker.io/fallenbagel/jellyseerr:preview-OIDC` branch which is slightly out of date. But hundreds of others are and I'm not going to be updating any time soon.


```
BASE="https://serrr.domain.com"   # <- change
EMAIL="vocoder@x.com"               # <- change
PASS="xxxxxxxxxxxxxx"                   # <- change

# 1) login; capture session cookies
curl -sk -c /tmp/seerr.cookies \
  -H 'Content-Type: application/json' \
  --data "{\"email\":\"$EMAIL\",\"password\":\"$PASS\"}" \
  "$BASE/api/v1/auth/local" | jq '{id,username,permissions}'

>> 
{
  "id": 13,
  "username": null,
  "permissions": 0
}


# 2) immediately ask server who you are, using same cookies
curl -sk -b /tmp/seerr.cookies \
  "$BASE/api/v1/auth/me" | jq '{id,username,permissions}'

>>

{
  "id": 13,
  "username": "vocoder",
  "permissions": 162
}

```

This already conflicts with what https://github.com/damontecres/Wholphin/blob/16ac02a3fd02ef96456d5c24b772503bec8c8981/app/src/main/seerr/seerr-api.yml#L10 is saying in my env. Checking the apk when its done building.

voc0der:
Upon testing it, Request is greyed out again. That is after 'successfully' logging in again on Wholphin.
 
https://github.com/voc0der/Wholphin/commit/5c701fc94075025d88162820bb448748912b700f

https://github.com/voc0der/Wholphin/releases/tag/develop

image

Seerr fork PR: https://github.com/seerr-team/seerr/discussions/1529

version | "develop-7f3979411655ef5f65ce4918fe0f8f214403f75e"
commitTag | "7f3979411655ef5f65ce4918fe0f8f214403f75e"
updateAvailable | true
commitsBehind | -1
restartRequired | false

It's quite possible it's the seerr version that is causing this.

I do have authelia as an IdP in front of it, but it is being fully bypassed. Still I have found that to sometimes interfere with websockets, but not usually cookies.

damontecres:
So the version you are using is still in development? All of the integration testing was done on Jellyseer `v2.7.3`. I've clarified this in #765.

I'm happy to consider simple solutions to help support older versions, but I don't plan to work on it myself. Any proposed changes will need to be thoroughly tested and fully explained by the author.

voc0der:
Fair enough. I will close this since it feels like what you want. Obviously I have no intention of 'fixing' my environment to have the worse version of Seerr. So, I have my own fork this and seerr now and I'll continue using them, since I can't choose when they'll rebase the PR for Seerr OIDC. If anyone stumbles on this thread and is in my boat, feel free to try my APK's or fork my repo then build yourself.

voc0der:
Circling back to this, I'm a little confused how this works for any client..

I re-based my OIDC fork of seerr into 2.7.3 myself and found it broken in (stock) Wholphin still (works otherwise). I started verifying the API routes and session handling and its identical between my fork and stock 2.7.3,

This is the `/local` route that we hit in Wholphin against seerr 2.7.3 stock code:
https://github.com/seerr-team/seerr/blob/v2.7.3/server/routes/auth.ts#L614

This handler doesn't return `username` or `permissions` and those 2 elements are required by Wholphin.  https://github.com/damontecres/Wholphin/pull/734#issuecomment-3783736130

If this is the case, perhaps the proposed solution I had given is correct for all clients. https://github.com/damontecres/Wholphin/pull/734/changes#diff-62202585ba125a8c469f1a75d60773201171db0b6b77339271e7ea97bc1fe22aR189

I'm defending this approach, because https://github.com/damontecres/Wholphin/blob/main/app/src/main/seerr/seerr-api.yml#L3927 the next logical passthru endpoint in seerr that does include both fields is `/auth/me`.

`checkUser` runs for all routes in auth.ts https://github.com/seerr-team/seerr/blob/v2.7.3/server/routes/index.ts#L47 but it won’t populate `req.user` on the first `/local` request, since the session is only set at the end of that handler. So unless the client calls `/auth/me` afterward (or `/local` is changed to select/return those fields), the client won’t see username/permissions.

Lastly, even the `/jellyfin` route needs this. It does include the `permissions` in all paths, but there are some where it can be missing the `username`. It seems the safe (and intended) solution.

Reopened the PR and subtracted what you've already implemented as a fix.

damontecres:
Thanks for pushing on this. I believe you are correct that this is needed for local login. I have done most of my testing of the actually functionality with a Jellyfin user. Unfortunately, I really only tested whether login worked for local users.

So with that in mind, I looked into the Jellyseerr code and while I'm not very familiar with their code nor ORM, I think most of my testing going through this [block of code](https://github.com/seerr-team/seerr/blob/v2.7.3/server/routes/auth.ts#L418-L442) which updates the Jellyseerr database for the Jellyfin user. I'm guessing the ORM `save` function returns the entire User object including the permissions field.

Your proposed change certainly fixes local user login. And should make Jellyfin user login more robust as well. Thanks for the contribution!

PR #569, opened 2025-12-24
Integrate with libass-android to support SSA/ASS subtitles in ExoPlayer

damontecres:
## Description
Add better ASS/SSA support to ExoPlayer by using [`libass-android`](https://github.com/peerless2012/libass-android).

This is enabled with the "Use libass for ASS subtitles" advanced settings for ExoPlayer. It is enabled by default.

### Related issues
Related to #22

PR #736, opened 2026-01-21
Add setting to automatically switch between ExoPlayer & MPV

### LINKED ISSUE #253: [FEA] - Option to allow automatic switching between ExoPlayer and MPV based on HDR or Audio
### Description

The new MPV option is fantastic (amazing work on the latest release!), particularly for us Anime fans who use a lot of ASS subtitles. As is known, however, [the upstream](https://github.com/mpv-android/mpv-android/issues/874) does not currently support native HDR nor audio passthrough. Until that happens, I propose an option in settings (suggested by the Reddit user Debisibusis) that, if set, allows for Wholphin to automatically choose between MPV or Exoplayer based on whether a video having HDR content or not, which would allow for the best of both worlds - direct playing of ASS subtitles + proper HDR.

Separately, there could also be a toggle to switch to Exoplayer if the audio track selected is such that lossiness would occur due to the lack of audio passthrough. I am not very versed in this area, however, so I am unaware of whether this one would be very complex or not.

damontecres:
## Description
The MPV backend is no longer considered experimental. This PR adds a setting for player backend "Prefer MPV" which uses MPV for all playback unless the video is HDR which then uses ExoPlayer.

Switching occurs per video. So if there's a playlist with both SDR & HDR, the player will switch back and forth as needed.

### Related issues
Closes #253
Related to #235

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/services/AppUpgradeHandler.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/services/AppUpgradeHandler.kt
@@ -2,21 +2,26 @@ package com.github.damontecres.wholphin.services
 
 import android.content.Context
 import android.os.Build
+import android.widget.Toast
 import androidx.core.content.edit
 import androidx.datastore.core.DataStore
 import androidx.preference.PreferenceManager
+import com.github.damontecres.wholphin.R
 import com.github.damontecres.wholphin.WholphinApplication
 import com.github.damontecres.wholphin.preferences.AppPreference
 import com.github.damontecres.wholphin.preferences.AppPreferences
+import com.github.damontecres.wholphin.preferences.PlayerBackend
 import com.github.damontecres.wholphin.preferences.update
 import com.github.damontecres.wholphin.preferences.updateAdvancedPreferences
 import com.github.damontecres.wholphin.preferences.updateInterfacePreferences
 import com.github.damontecres.wholphin.preferences.updateLiveTvPreferences
 import com.github.damontecres.wholphin.preferences.updateMpvOptions
 import com.github.damontecres.wholphin.preferences.updatePlaybackOverrides
+import com.github.damontecres.wholphin.preferences.updatePlaybackPreferences
 import com.github.damontecres.wholphin.preferences.updateSubtitlePreferences
 import com.github.damontecres.wholphin.ui.preferences.PreferencesViewModel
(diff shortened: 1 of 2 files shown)


zoro11031:
Amazing work. ~~Will this setting be enabled by default?~~ (never mind, I looked at the code and can see that it is. Awesome!)

damontecres:
> Will this setting be enabled by default?

Yes, it is the default for new installs. Still debating if I should change to it when users upgrade or not. I think I might and have it show a toast.

enormouspoon:
This feature is absolutely amazing. I was trying to find a way to do this without switching back and forth manually. I was researching if there was a way to pick the player at the library level (MPV for Anime/SSA subs, Exo for everything else) but this feature sounds even better. Thank you so much. Look forward to it being implemented. 

[edit] just donated because thank you.

mjnewton:
Reading between the lines, I am assuming that using MPV is preferential to Exoplayer?

I tried Googling to see if I could determine this for myself but got quite a mixed bag about the pros and cons, mostly focusing on subtitle support which doesn't happen to be something I need.

If someone could give an overview or provide some pointers I'd welcome it. In the meantime I'll switch over to MPV, even if I'll have a nagging thought at the back of my mind wondering what I'm getting! :-)

damontecres:
> If someone could give an overview or provide some pointers

Just a quick pro/con list. Wholphin's MPV implementation is still a bit more buggy than I thought (eg #755, #766), so make "Prefer MPV" the default may be reverted before the next release while I continue to work on it.

### MPV
 
#### Pros
- Direct play pretty much everything
- Significantly better subtitle support, particular ASS (#569 will close the gap though)
- Easy subtitle offset adjustments
- Huge amount of [configuration options](https://mpv.io/manual/stable/#options) to control many aspects of playback

#### Cons
- Cannot play HDR content on Android
- Always decodes audio and outputs PCM
    - So older TV without eARC will be limited to 5.1
    - Loss of "spatial" audio
- Can be temperamental, Wholphin's implementation is probably more prone to bugs than ExoPlayer


### ExoPlayer

#### Pros
- Supports HDR playback
- Supports audio passthrough
- Support DRM content
- Battle tested & backed by Google

#### Cons
- Doesn't support many codecs or containers, especially older or proprietary ones, such as avi, divx, wmv
- ASS subtitle support is limited (#569 will help greatly)

PR #738, opened 2026-01-21
Added wholphin://play Intent

sysmoon14:
## Description
This PR implements **new deep linking functionality** for media playback, enabling external tools (e.g., ADB, Home Assistant) to trigger content directly.

Previously, there was no support for deep linking directly into playback. This PR introduces the `wholphin://play/` scheme and adds specific intent parameters (`userId`, `serverId`) to **bypass the login and server selection screens entirely**, allowing for seamless automation from a cold or warm start.

**Changes Implemented:**
* **New Deep Link Support:** Added handling for `wholphin://play/{itemId}` URIs.
* **Auth Bypass for Automation:** Implemented logic to accept `userId` and `serverId` via intent extras.
    * If these IDs are present, the app skips the standard startup flow (Server/User Select) and forces a session restoration for the requested user.
    * This works even if the app is already running (via `onNewIntent`), allowing for instant user switching.
* **Robust State Handling:**
    * **Cold Start:** App initializes, logs in silently, and starts playback.
    * **Warm Start / Wrong User:** If the app is already open but on a setup screen or logged into a different user, the activity is restarted to ensure a clean authentication flow.
    * **Same User:** If the app is already logged in as the target user, navigation happens instantly without reloading.
* **UUID Helper:** Added robust parsing to handle UUIDs with or without dashes (common in external tools).
* **Duplicate Prevention:** Added logic to ignore the playback command if the requested item is already playing.

### Related issues
Implements deep linking support for home automation integration.

**Steps to verify:**
1.  Close the app (or leave it on the User Select screen).
2.  Send the following ADB command:
    ```bash
    adb shell am start -W -a android.intent.action.VIEW \
    -d "wholphin://play/{ITEM_UUID}?autoplay=true" \
    --es "userId" "{USER_UUID}" \
    --es "serverId" "{SERVER_UUID}" \
    -n com.github.damontecres.wholphin/.MainActivity
    ```
3.  Observe that the app launches, bypasses all selection screens, and immediately starts playing the content.

### Screenshots
N/A - Logic implementation only.

### AI/LLM usage
I utilized an LLM (Gemini) to assist with the architecture of the `appStart` authentication flow and the intent handling lifecycle (`onNewIntent`) to ensuring smooth transitions between cold and warm states. Validated via ADB debugging and Home Assistant integration.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
@@ -2,6 +2,7 @@ package com.github.damontecres.wholphin
 
 import android.content.Intent
 import android.content.res.Configuration
+import android.net.Uri
 import android.os.Bundle
 import android.view.WindowManager
 import androidx.activity.compose.setContent
@@ -73,6 +74,7 @@ import okhttp3.OkHttpClient
 import org.jellyfin.sdk.model.api.BaseItemKind
 import org.jellyfin.sdk.model.serializer.toUUIDOrNull
 import timber.log.Timber
+import java.util.UUID
 import javax.inject.Inject
 
 @AndroidEntryPoint
@@ -149,7 +151,17 @@ class MainActivity : AppCompatActivity() {
                 window?.clearFlags(WindowManager.LayoutParams.FLAG_SECURE)
             }
         }
-        viewModel.appStart()
+        
+        // PARSE IDs ROBUSTLY
+        val overrideUserId = parseUUID(intent?.getStringExtra(INTENT_USER_ID))
+        val overrideServerId = parseUUID(intent?.getStringExtra(INTENT_SERVER_ID))
+
+        // Capture requested destination from intent in onCreate so it's available when AppContent
+        // is composed (avoids losing item/play target after nuclear restart or late composition).
+        viewModel.pendingRequestedDestination = extractDestination(intent)
+
+        viewModel.appStart(overrideUserId, overrideServerId)
+        
         setContent {
             val appPreferences by userPreferencesDataStore.data.collectAsState(null)
             appPreferences?.let { appPreferences ->
@@ -232,6 +244,7 @@ class MainActivity : AppCompatActivity() {
                                                                     appPreferences.updateUrl,
                                                                 )
                                                             } catch (ex: Exception) {
+                                                                if (ex is kotlinx.coroutines.CancellationException) throw ex
                                                                 Timber.w(
                                                                     ex,
                                                                     "Exception during update check",
@@ -257,9 +270,9 @@ class MainActivity : AppCompatActivity() {
 
                                                     if (showContent) {
                                                         val requestedDestination =
-                                                            remember(intent) {
-                                                                intent?.let(::extractDestination)
-                                                            }
+                                                            viewModel.pendingRequestedDestination
+                                                                ?.also { viewModel.pendingRequestedDestination = null }
+                                                                ?: intent?.let(::extractDestination)
                                                         ApplicationContent(
                                                             user = current.user,
                                                             server = current.server,
@@ -305,7 +318,9 @@ class MainActivity : AppCompatActivity() {
     override fun onRestart() {
         super.onRestart()
         Timber.d("onRestart")
-        viewModel.appStart()
+        val overrideUserId = parseUUID(intent?.getStringExtra(INTENT_USER_ID))
+        val overrideServerId = parseUUID(intent?.getStringExtra(INTENT_SERVER_ID))
+        viewModel.appStart(overrideUserId, overrideServerId)
     }
 
     override fun onStop() {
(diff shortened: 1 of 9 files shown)


damontecres:
Hi, thanks for the contribution! Can you elaborate on the use case for this PR?

Is there already a defined standard (maybe just for arguments?) for starting app playback in Home Assistant? I know the Jellyfin server has websocket commands for browsing and playing media that Wholphin does not implement yet.

sysmoon14:
Hi, yes of course.

Home assistant has an integration with Jellyfin already, however it can only initiate playback to a client device if the client is already in the Jellyfin app and waiting for a command.

The integration is useful for managing playback media that has already started, or state based automations (e.g. dim the lights when a movie starts), but it's not good for initiating playback to begin with unless you're already in the app.

Enter android intents, this allows me to use the Home Assistant "Android Debug Bridge" to send ADB commands to my TV device.

In my case I have some NFC cards for my little girl that she can tap on a sensor, doing so starts the movie printed on the card. The official Jellyfin and Plex clients support android intents that will start their respective apps and navigate to the media item ID provided.

The limitation with those clients is that they both stop at the "User Select" screen if you have multiple users, and even if you don't, they require an "Ok/Enter" command to be sent to press play on the item it navigated to.

This PR takes it a step further by adding support for autoplay, sending a user ID and sending a server ID, meaning the command can be a one shot!

Some links for context below 

https://github.com/jellyfin/jellyfin-androidtv/discussions/3452#discussioncomment-15552168

https://simplyexplained.com/blog/how-i-built-an-nfc-movie-library-for-my-kids/

voc0der:
Adding that I am willing to help test this. I use HomeAssistant for this exact purpose. (in my case, I have a voice assistant I can ask to play a movie/ TV show), etc.

Hukuma1:
Would love to test this so that I can integrate a button into Home Assistant to quickly load specific content. Would this work with playlists too? Could it also do a randomize sort perhaps? (e.g. load random cartoon from playlist)

app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
```diff
@@ -346,22 +425,261 @@ class MainActivity : AppCompatActivity() {
...
+                    seasonId = intentData.seasonId,
+                    seasonNumber = intentData.seasonNumber,
+                    episodeId = itemId,
+                    episodeNumber = intentData.episodeNumber,
+                ),
+            )
+        }
+        
+        if (autoplay) {
+            val currentUser = viewModel.serverRepository.currentUser.value
```
damontecres: Some of this logic is already implemented in `SeriesViewModel`, so would be good reuse that maybe by creating a service class.

app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
```diff
@@ -377,20 +695,80 @@ class MainActivity : AppCompatActivity() {
...
+        
+        return if (type != null) {
+            Destination.MediaItem(itemId = itemId, type = type, autoPlayOnLoad = autoplay)
+        } else {
+            Timber.d("Cannot create destination for $itemId: type unknown and autoplay=false")
+            null
+        }
+    }
+    
+    private fun parseUUID(input: String?): UUID? {
```
damontecres: `String.toUUIDOrNull()` from the jellyfin sdk already handles this.

app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
```diff
@@ -404,23 +782,52 @@ class MainActivityViewModel
         private val deviceProfileService: DeviceProfileService,
         private val backdropService: BackdropService,
     ) : ViewModel() {
-        fun appStart() {
+
+        var pendingRequestedDestination: Destination? = null
+        var lastProcessedItemId: UUID? = null
+        var pendingIntentData: MainActivity.IntentData? = null
```
damontecres: Should be using state or passing these around in functions. Just parking variables in the ViewModel isn't great.

app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
```diff
@@ -232,6 +304,7 @@ class MainActivity : AppCompatActivity() {
                                                                     appPreferences.updateUrl,
                                                                 )
                                                             } catch (ex: Exception) {
+                                                                if (ex is kotlinx.coroutines.CancellationException) throw ex
```
damontecres: What is this for?

damontecres (changes requested):
There's a lot of hard to follow code here, so it needs some TLC before it can be considered.

damontecres:
Closing due to inactivity. Feel free to update the PR and re-open it.

Hukuma1:
Any chance you could please take a look at this addition personally @damontecres? Unfortunately the OP of this PR is MIA, but the functionality addition opens up great possibilities. (using Home Assistant/ADB to launch specific content is very powerful). I know the official client already supports deep linking, could we at least just mirror it? Can the code be transferred?

damontecres:
Opened https://github.com/damontecres/Wholphin/pull/1688 to replace this PR

PR #753, opened 2026-01-23
Skip Server Select Screen When Only One Server Exists

sysmoon14:
When "Sign in automatically" is disabled and only one server is configured, automatically skip the server select screen and navigate directly to the user select screen. Removed intermediate Loading screen navigation to prevent UI flash.

## Description

This PR improves the user experience for users with a single server configuration. When "Sign in automatically" is disabled and only one server exists, the app now automatically skips the server select screen and navigates directly to the user select screen for that server.

Additionally, the intermediate `Loading` screen navigation was removed to prevent a brief flash of the server select screen before redirecting to the user select screen.

### Changes
- Modified `appStart()` in `MainActivityViewModel` to check if only one server exists when "Sign in automatically" is disabled
- If a single server is found, navigate directly to `SetupDestination.UserList` instead of `SetupDestination.ServerList`
- Removed the intermediate `SetupDestination.Loading` navigation to prevent UI flash
- Simplified `appStart()` function signature by removing unused override parameters

### AI/LLM usage
Used AI assistance to implement the initial feature and refine the logic based on feedback. The code was tested manually to verify:
- Single server configuration correctly skips server select screen
- Multiple server configuration still shows server select screen
- No UI flash when navigating to user select screen
- Behavior only applies when "Sign in automatically" is disabled

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/data/model/BaseItem.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/data/model/BaseItem.kt
@@ -6,6 +6,7 @@ import androidx.compose.runtime.Stable
 import androidx.compose.ui.text.AnnotatedString
 import androidx.compose.ui.text.buildAnnotatedString
 import com.github.damontecres.wholphin.ui.DateFormatter
+import com.github.damontecres.wholphin.ui.abbreviateNumber
 import com.github.damontecres.wholphin.ui.detail.CardGridItem
 import com.github.damontecres.wholphin.ui.detail.series.SeasonEpisodeIds
 import com.github.damontecres.wholphin.ui.dot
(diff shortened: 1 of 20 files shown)


damontecres:
If "Sign in automatically" is disabled, then the app should show the most recently used server user list when starting.

So I'm not sure this PR is necessary.

damontecres:
Closing for inactivity, feel free to re-open to continue the discussion.

sysmoon14:
You're right that the server select screen shows the most recent server. But if there's only one servery why show it at all? This PR just saved a click is all.

sysmoon14:
@damontecres - I believe there is some value to this. You're right that the app normally shows the user list of the most recent server, but sometimes (seems reproduceable by closing the app with continued back button presses) it shows the server list.

PR #761, opened 2026-01-24
Fix(player): add exoplayer fallback for all 4k playback  when MPV software decoding is enabled

zoro11031:
## Description
Updates the  logic in the Prefer MPV setting to prevent performance issues when software decoding is selected. When the MPV backend is configured to use software decoding, the app will now automatically fallback to ExoPlayer for 4K content. Most Android TV devices cannot smoothly render 4K video via software decoding, so falling back to Exoplayer for this content makes sense in most cases when using the software renderer. This mirrors the existing fallback mechanism used for HDR content.

### Related issues
N/A

### Screenshots
N/A

### AI/LLM usage
None


damontecres (commented):
I think this is a good idea

damontecres (approved):

Issue #735, opened 2026-01-21
[FEA] - Make fonts under people circles smaller; make role smaller weight font than name? Same for movie + year?


Pentaphon:
### Description

I'm all about readability, but now that the people photos are circles, the font under them seems too awkwardly big now. I was looking at Plex and they use smaller fonts in general but for circles, they use a small font for the actor's name and a _**slightly smaller weight**_ font for the role. I think this is a good way to go to make them more distinct.

Also, this same style could be used for movie posters where the title is a regular weight but the YEAR under it is a slightly smaller weight. In general, some font size and weight tweaking is needed for the text under both circles and posters. The movie posters might only need the year to be a smaller weight to add some extra distinction. 

Thoughts? See screenshot below for those reading that aren't familiar with font weights. If an actor name is 400 weight, the role would be 300, etc. 

<img width="925" height="438" alt="Image" src="https://github.com/user-attachments/assets/d6902d0f-86d8-4a24-842d-421cc2819179" />

Pentaphon:
@YogiBear12 do you have any thoughts on my issue? I was just looking at your typography PR and am interested if any feedback you might have, if you have any free time.

YogiBear12:
@Pentaphon I like your suggestion of separating actors from the roles by using a slightly different font weight. I'll need to review what's currently there, but I can certainly try to touch them up

damontecres:
I think this is a good idea. It's probably just a one line change to add a font weight [here](https://github.com/damontecres/Wholphin/blob/main/app/src/main/java/com/github/damontecres/wholphin/ui/cards/PersonCard.kt#L188).

YogiBear12:
@damontecres @Pentaphon Played around with it a bit.
Used bodyMedium with semiBold for the name, and bodySmall with normal for the role. 

<img width="960" height="540" alt="Image" src="https://github.com/user-attachments/assets/2d169ab3-b0ee-4f2b-83c8-f898b862065d" />

Pentaphon:
@YogiBear12 that looks so much better with the circles. They needed a smaller font badly, and it looks more like Plex without being too small to read.  

Do you think you could apply the same or similar for the posters (title+year) to make everything consistent? The year especially needs more distinction for posters.

YogiBear12:
I can do something similar for the library view. Just need to test how well it scales when you change your number of columns or spacing

YogiBear12:
[...]

PR #775, opened 2026-01-26
Title/subtitle style update

YogiBear12:
## Description
This PR updates cards that have a title and subtitle, so that they are visually distinct instead of sharing the same font and style.
Titles use bodyMedium typography with semiBold weight. 
Subtitles use bodySmall typography with normal weight.

This affects:
- PersonCard
- EpisodeCard
- DiscoverItemCard
- GridCard
- SeasonCard

### Related issues
Related to [735](https://github.com/damontecres/Wholphin/issues/735)

### Screenshots
Screenshot_20260125_190057

### AI/LLM usage
No LLM usage

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/cards/DiscoverItemCard.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/cards/DiscoverItemCard.kt
@@ -27,6 +27,7 @@ import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.res.pluralStringResource
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.unit.Dp
 import androidx.compose.ui.unit.dp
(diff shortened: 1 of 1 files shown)


damontecres:
I think this makes the text too small. I think increasing to bodyLarge/bodyMedium would be good.

YogiBear12:
@damontecres Here's a poorly photoshopped example for comparison so you can see how it looks against the existing text.
Happy to go with whatever - I can take another stab at it this weekend using bodyLarge and bodyMedium. I just chose these options because part of the feedback in [735](https://github.com/damontecres/Wholphin/issues/735) was that they needed to be a tad smaller.

540211113-2d169ab3-b0ee-4f2b-83c8-f898b862065d (1)

YogiBear12:
@damontecres 

Okay I finally got around to this and have attached screenshots for the three styles (current, medium/small and large/medium). I'm still partial to medium/small, at least on the PersonCard. But I think the title/subtitle styling should be consistent across all cards, and I'm not opposed to how large/medium looks on the other cards

PersonCard current state
PersonCard standard

PersonCard medium/small
PersonCard medium-small

PersonCard large/medium
PersonCard large-medium

GridCard current state
GridCard standard

GridCard medium/small
GridCard medium-small

GridCard large/medium
GridCard large-medium

SeasonCard current state
SeasonCard standard

SeasonCard medium/small
SeasonCard medium-small

SeasonCard large/medium
SeasonCard large-medium

Pentaphon:
> I think Medium/Small is good.

Agreed!

PR #780, opened 2026-01-26
Fix: Prevent password from being sent as API key header

voc0der:
## Summary
Fixes an issue where passwords for local/Jellyfin authentication were incorrectly being sent as API key headers.

## Changes
- Only send `X-Api-Key` header when using API key authentication
- For session-based auth (local and Jellyfin), avoid treating password as API key

## Technical Details
The `passwordOrApiKey` parameter is now conditionally used based on the authentication method. Only when `authMethod == SeerrAuthMethod.API_KEY` should it be passed to the `SeerrApiClient` constructor as an API key.

## Related issues
Replaces #734 

## AI/LLM usage
Claude + ChatGPT usage


damontecres:
Implemented in #783

Issue #522, opened 2025-12-20
Advance by holding left/right


Shamatix:
### Description

Currently you can tap, to skip x seconds forward or backwards. However if you want to scrub backwards/forward you can't simply hold the left/right button down to do so, currently you first need to select the "timeline" before being able to do so.

I suggest the following:
1. Tap to skip (predefined values in settings - As we currently have) 
2. If you hold left/right it should bring up player UI and scrub back/forward (fast forward/backwards as Netflix hbo etc does it).

Pentaphon:
@Shamatix I'm sure the dev can add an option to "advance by holding left/right" in the settings to change to this desired behavior. I'm sure you're not the only user out there who would like this. You should change the title of this issue to something like "Please add optional hold to advance controls in player". 

Or perhaps more simply, the default player control behavior can be changed to:

1. Tap to skip (predefined values in settings)
2. Hold left/right to bring up player UI (or just the progress bar?) and scrub back/forward. 

@damontecres thoughts on what can be done to appease users who just want to quickly "hold left/right to scrub forward" similar to how we can now quickly tap the center button to pause?

Shamatix:
> [@Shamatix](https://github.com/Shamatix) I'm sure the dev can add an option to "advance by holding left/right" in the settings to change to this desired behavior. I'm sure you're not the only user out there who would like this. You should change the title of this issue to something like "Please add optional hold to advance controls in player".
> 
> Or perhaps more simply, the default player control behavior can be changed to:
> 
> 1. Tap to skip (predefined values in settings)
> 2. Hold left/right to bring up player UI and scrub back/forward.
> 
> [@damontecres](https://github.com/damontecres) thoughts on what can be done to appease users who just want to quickly "hold left/right to scrub forward" similar to how we can now quickly tap the center button to pause?

Thank you for your comment, as you might have noticed, english is far from my first language ^^ I have now changed the text and title a bit following your suggestion:)

Pentaphon:
@Shamatix @mjnewton Since you both requested this and I agree this is a useful feature...any thoughts on: 

1. What kind of UI that should show up when advancing? I was thinking just showing the progress bar and nothing else. A minimal version of the full player UI, if you will.
2. How fast to advance when holding? Should it be a smooth advance that stays the same no matter how long you hold or an accelerating advance that gets faster the longer you hold it? I'm thinking a smooth advance would be enough. 

[...]

PR #784, opened 2026-01-27
Add seek acceleration when holding direction keys

voc0der:
## Summary
Adds progressive seek acceleration when holding left/right direction keys in the playback seek bar.
This leaves the default +15/30/-15/30 seek feature untouched, it's only if you're on the seek bar above that (press up).
This change extends `BasePlayer`, `PlayBackControls` takes `Player` and both mpv and ExoPlayer use `IntervalSeekBarImpl`, so should work for both. Tested it on ExoPlayer a bit, and it works great.

## Changes
- Tracks key repeat count for left/right holds
- Applies speed multipliers: 1x (0-15), 2x (15-30), 3x (30-45), 4x (45+)
- Works for both forward and backward seeking
- Resets multiplier on key release

## Behavior
Traditional media player UX: the longer you hold a seek direction, the faster it scrolls through the timeline.

## Related Issues
Related to #522

## AI/LLM usage
Used Claude 4.5 Sonnet for most of the changes and to help write this PR. But both are second checked/modified by me.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/playback/SeekBar.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/playback/SeekBar.kt
@@ -80,15 +80,16 @@ fun SteppedSeekBarImpl(
         enabled = enabled,
         progress = progressToUse,
         bufferedProgress = bufferedProgress,
+        durationMs = durationMs,
         onLeft = { multiplier ->
             controllerViewState.pulseControls()
-            seekProgress = (progressToUse - offset * multiplier).coerceAtLeast(0f)
+            seekProgress = (seekProgress - offset * multiplier).coerceAtLeast(0f)
             hasSeeked = true
             seek(seekProgress)
         },
         onRight = { multiplier ->
             controllerViewState.pulseControls()
-            seekProgress = (progressToUse + offset * multiplier).coerceAtMost(1f)
+            seekProgress = (seekProgress + offset * multiplier).coerceAtMost(1f)
             hasSeeked = true
             seek(seekProgress)
         },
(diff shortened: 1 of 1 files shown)


voc0der:
Testing a variation of this that is smarter; adapting with the video length, since the experience can vary depending on the timeline length. Stay tuned.

voc0der:
This version is solid. I've tested it against videos of varying lengths, and the UX is way better than without.

Pentaphon:
@voc0der when you hold the direction key, does it bring up the entire player UI or just the scrollbar? I think it would be neat if it just brought up the scrollbar to use minimal UI, similar to how a single left or right press just brings up a single progress icon. Please consider that, if possible.

mjnewton:
> @voc0der when you hold the direction key, does it bring up the entire player UI or just the scrollbar? I think it would be neat if it just brought up the scrollbar to use minimal UI, similar to how a single left or right press just brings up a single progress icon. Please consider that, if possible.

I was wondering similar, but re-reading the description summary it sounds like this press-and-hold functionality only takes effect if you're already on the seek bar  i.e. the UI (including seek bar) is already up?

Related to this, even aside from what is shown whilst seeking, I'd personally like the press-and-hold functionality to apply whenever I press the button regardless where I am (as per https://github.com/damontecres/Wholphin/issues/522). Having to first press up (or down) wouldn't feel intuitive and indeed would seem an unnecessary extra step.

voc0der:
Hello, I was hoping to keep the discussion related to that here so it doesn't slow down implementing this PR:

https://github.com/damontecres/Wholphin/issues/522#issuecomment-3806432970

This PR addresses part one (imo) which is to make the seek bar work better when holding directions down when already on the timeline. It does not change tap/hold to move you onto it. 

Still accepting & thinking about ideas and need to see exactly how it currently works in that respect, and will make a separate PR for that, assuming the maintainer agrees.

voc0der:
Closing in favor of #900

Issue #810, opened 2026-02-01
[FEA] - Hide spoilers for unwatched items


skoshy:
### Description

It would be nice to hide spoilers for unwatched items in our library. Specifically for the movie/episode's description, but could apply to the thumbnail on an episode too.

Nocifer:
Duplicate of #360 

Seeing as you're already at it, could you perhaps make the setting more granular (i.e. a separate toggle for movies, shows and episodes) as described in that other report?

skoshy:
Done! I'll close this issue in favor of https://github.com/damontecres/Wholphin/issues/360

Issue #360, opened 2025-12-02
[FEA] - Option to hide spoilers (i.e. plot summaries) for unwatched media


Nocifer:
### Description

Just a minor feature request for down the road; as the title says, I'd like the ability to hide the plot summaries for any unwatched movies and/or shows and/or episodes (the 'and/or' is important, this should be configurable per type, not a catch-all on/off switch).

This is something that Jellyfin's Kodi addon does and it's a great little detail for those of use who absolutely hate being spoiled by summaries.

MorningStarGG:
Been trying to find a client that would handle this as well

Pentaphon:
> I'd like the ability to hide the plot summaries for any unwatched movies and/or shows and/or episodes

I don't really see "spoilers" in the metadata from tmdb. The people curating that info tend to do a good job not spoiling the movie or episode.

MorningStarGG:
> > I'd like the ability to hide the plot summaries for any unwatched movies and/or shows and/or episodes
> 
> I don't really see "spoilers" in the metadata from tmdb. The people curating that info tend to do a good job not spoiling the movie or episode.

Perhaps, but some don't want to see any information on the episode or see a thumbnail/image from it unless it's been watched. Not everyone uses TMDB. Blurring the images of unwatched episodes and blurring/hiding the episode summary/description would be fairly simple and would help a lot of people.

Personally, I want this as a lot of thumbnails are spoilers, and descriptions, even vague ones, could contain too much.

xtrandom:
@damontecres I hope you're doing well. Any chance you could rework the closed PR (https://github.com/damontecres/Wholphin/pull/811) to add this functionality please?

PR #811, opened 2026-02-01
Add spoiler-related options - hide unwatched item titles, overviews, and thumbnails

skoshy:
## Description

To prevent spoilers, I wanted to add an option to hide unwatched item overviews / descriptions. I also added the ability to hide unwatched episode thumbnails, either with a blur effect or by using the series thumbnail.

This also addresses items without an overview by giving them a "No overview" text.

I intentionally left the overview still visible when viewing an item's media information, so the user can still look at the overview if they really intend to dig deeper into the item.

### Related issues

https://github.com/damontecres/Wholphin/issues/810
https://github.com/damontecres/Wholphin/issues/360

### Screenshots

Menu items:

Screenshot_20260201_181237

Hidden overview, blurred thumbnail:

Screenshot_20260201_181300

Hidden overview, hidden title, series thumbnail:

Screenshot_20260201_181222

### AI/LLM usage

I used Gemini to help with this PR. I've tested the code manually in Android Studio (how I generated the screenshots).

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/data/model/BaseItem.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/data/model/BaseItem.kt
@@ -5,6 +5,7 @@ import androidx.compose.runtime.Immutable
 import androidx.compose.runtime.Stable
 import androidx.compose.ui.text.AnnotatedString
 import androidx.compose.ui.text.buildAnnotatedString
+import com.github.damontecres.wholphin.preferences.UserPreferences
 import com.github.damontecres.wholphin.ui.DateFormatter
 import com.github.damontecres.wholphin.ui.abbreviateNumber
 import com.github.damontecres.wholphin.ui.detail.CardGridItem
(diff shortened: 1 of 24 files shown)


MorningStarGG:
So you're also working on one to blur/hide the thumbnails? Either way, this is a good step towards solving one of the more major missing things that most clients ignore.

skoshy:
I've updated the PR, there's now more granular control for movies, episodes, and also episode titles which can also be a source of spoilers.

I've also added the ability to hide thumbnails for episodes - either by blurring them or by using the series thumbnail

MorningStarGG:
I have built and tested this on the latest git as of today's 0.4.2 and this is working on my onn 4k Pro Android 14. However is the title being hidden also supposed to hide the title of the episode in the player? Right now it doesn't. So if I bring up the media controls the title still shows even with the option to hide them enabled.

Also still shows the thumbnail, episode info, and title in the next up box at the end of watching an episode. I have title, description hidden and thumbnail blur on but seems to not apply to the next up for some reason.

Otherwise, seems good here.

app/src/main/java/com/github/damontecres/wholphin/ui/cards/BannerCard.kt
```diff
@@ -109,13 +129,17 @@ fun BannerCard(
                     contentDescription = null,
                     contentScale = ContentScale.FillBounds,
                     onError = { imageError = true },
-                    modifier = Modifier.fillMaxSize(),
+                    modifier =
+                        Modifier
+                            .fillMaxSize()
+                            .dimAndBlur(finalBlur),
```
damontecres: The blur effect does not work for Android before 12, so a lot of the features implemented in this PR won't work for many users.

app/src/main/java/com/github/damontecres/wholphin/ui/cards/ItemCardImage.kt
```diff
@@ -63,15 +65,32 @@ fun ItemCardImage(
...
+    val isEpisodeThumbnail =
+        isEpisode &&
+            !item.useSeriesForPrimary &&
+            imageType == ImageType.PRIMARY
+
+    val useSeriesThumbnail =
+        isEpisodeThumbnail && isUnwatched && spoilerMode == EpisodeThumbnailSpoilerMode.SPOILER_SERIES_THUMBNAIL
+    val finalBlur =
+        isEpisodeThumbnail && isUnwatched && spoilerMode == EpisodeThumbnailSpoilerMode.SPOILER_BLUR
+
```
damontecres: This composable really shouldn't have logic to it. Callers further up the chain should be telling it what to do.

app/src/main/java/com/github/damontecres/wholphin/data/model/BaseItem.kt
```diff
@@ -183,6 +184,41 @@ data class BaseItem(
         return result
     }
 
+    fun shouldHideOverview(preferences: UserPreferences): Boolean {
```
damontecres: This is a weird place to implement this logic.

Maybe move the spoiler settings into their own object and add extension functions to that for this logic.

damontecres (changes requested):
A good start, but needs some work.

skoshy:
Thanks! Have some other things going on at the moment, but I'll be going through these and making the necessary changes hopefully soon.

damontecres:
Closing as out dated for now. Feel free to update it & re-open

Issue #669, opened 2026-01-11
[FEA] - Change app language option


Generator:
### Description

Option to change app language instead of using system language.

Not everyone wants to use the same language of the system, also useful for issue reports.

damontecres:
This is similar to #513

Issue #513, opened 2025-12-20
[FEA] - Show different UI language for different users


Rand0mB0y:
### Description

My request is kindly the following...
While on the same Android TV box and without changing the Android language settings:
- If user 1 is logging in, show Wholphin in French
- If user 2 is logging in, show Wholphin in English
  
Jellyfin already has a "display language" setting in the "display" section of each users. Once set, this is supposed to be a global setting.
Any web-wrapped app is respecting this, but most native apps don't.
So, ideally, this would be handled automatically by following the admin setting for each users. 
But if retrieving that server setting is not an option, it would be already amazing if there could be a chance to have a Wholphin internal language setting for each user.
  
I live in a multi language household, sometimes with old folks coming visiting us, too, and they don't handle English too well. So this would help tremendously. 
  
Once again... a massive thank you for all your hard work! Every day this app becomes prettier and more refined!

oyvhov:
This would be a great addition to be able select Language. I have translated my language but unable to see it because Android wont let me set that variant of the language

Rand0mB0y:
@damontecres Thank you so very much for this! I'm loving it!!!!! It works perfectly

PR #831, opened 2026-02-05
feat: Add in-app language selection

XuanMinh1223:
## Description
This commit introduces an in-app language selection setting, allowing users to choose their preferred language independently of the system language.

Changes include:
*   Adding an `AppLanguages` enum to the `WholphinDataStore.proto` to persist the language choice.
*   Implementing the necessary UI in the Interface preferences screen, including a list of available languages (currently languages with translation of 90% or better).
*   Adding logic to prompt the user to restart the app after changing the language for the setting to take effect.
*   Updating `MainActivity` to apply the selected locale upon application startup.

### Related issues
#513 , #669 

### Testing
Tested on Android TV emulator, checked that changing language refreshed the app (if confirmed) and shows the correct language. Ensured language selection was persistent after closing and re-opening the app. Ensured that "Default" would be the option shown if user hadn't selected a language.

## Screenshots
Screenshot_1770267296
Screenshot_1770267305
Screenshot_1770267739

## AI or LLM usage
Used Android Studio's Gemini.  Gemini helped me decipher how settings were used in order to show how to best implement a language selector. Also used it to create the commit message.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
@@ -453,6 +453,32 @@ class MainActivity : AppCompatActivity() {
         }
     }
 
+    private fun applyAppLocale() {
+        val languageEnum = runBlocking {
+            userPreferencesDataStore.data.first().interfacePreferences.language
+        }
+        val languageTag = when (languageEnum) {
+            AppLanguages.ENGLISH -> "en"
+            AppLanguages.INDONESIAN -> "in"
+            AppLanguages.CZECH -> "cs"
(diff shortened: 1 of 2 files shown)


damontecres:
Thanks for the contribution. Unfortunately, it looks like you have a bad merge here that makes it impossible to review.

That said, using an enum for the language is not the right approach since there are many more languages than the 10 listed.

If you'd like to make a future contribution, please be sure to read the [Developer's guide](https://github.com/damontecres/Wholphin/blob/main/DEVELOPMENT.md) and the [Contributing details](https://github.com/damontecres/Wholphin/blob/main/CONTRIBUTING.md), especially the AI/LLM policy.

PR #834, opened 2026-02-05
Fix: suggestions logic/re-implement logic into worker and address performance issues

### LINKED ISSUE #833: [BUG] - Suggestions service fixes/follow ups
### Description

- I am seeing a performance regression when loading the home page after cold starting the app. Currently investigating.
-  When I moved/offloaded the suggestions logic from `SuggestionService.kt` to `SuggestionsWorker`, I re-implemented the actual seed items logic, but neglected to add the genre based suggestions into the final mix of suggestions. This is a simple fix

### Reproduction steps

_No response_

### App Version

 v0.4.1-9-gaeaecc0 

### Server Version

10.11.5

### Device

NVIDIA Shield TV Pro 2019

zoro11031:
## Description
Addresses suggestion logic oversight in recently merged suggestions fix, and improve performance (especially app homepage startup time)

- Re-implements the logic that sorts/categorizes tailored suggestions based on the seed items that are queried. This was implemented in the original PR at first, but I overlooked moving that logic
- Removes  `OneTimeWorkRequestBuilder` and replaces with trigger in `PeriodicWorkRequest` to fire with a 30 seconds delay if cache is empty.

 
### Related issues
Fixes #833

### Testing
Will be testing in:
Android TV emulator in Android Studio
NVIDIA Shield TV Pro 2019

## AI or LLM usage
I will use AI to help with writing the tests/testing services.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/services/SuggestionsSchedulerService.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/services/SuggestionsSchedulerService.kt
@@ -7,7 +7,6 @@ import androidx.work.BackoffPolicy
 import androidx.work.Constraints
 import androidx.work.ExistingPeriodicWorkPolicy
 import androidx.work.NetworkType
-import androidx.work.OneTimeWorkRequestBuilder
 import androidx.work.PeriodicWorkRequestBuilder
 import androidx.work.WorkManager
 import androidx.work.workDataOf
@@ -23,6 +22,7 @@ import java.util.UUID
 import javax.inject.Inject
 import kotlin.time.Duration.Companion.hours
 import kotlin.time.Duration.Companion.minutes
+import kotlin.time.Duration.Companion.seconds
 import kotlin.time.toJavaDuration
 
 @ActivityScoped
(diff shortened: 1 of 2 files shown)


zoro11031:
@damontecres I'm just throwing stuff at the wall in here trial and error style. Haven't had a chance to try this on hardware yet but I saw an improvement in the emulator. Not sure how that will translate to real world performance tho.

zoro11031:
I'm starting to wonder if the performance fixes are even necessary because the start up time issue has been non existent for me today.

damontecres:
The slow start up issue seems to only happen the first time if there are no suggestions cached. So I seem to hit it a lot on my emulator, but not so much on my actual devices.

Try adding an initial delay to the `OneTimeWorkRequestBuilder`. Maybe even just 30-60 seconds would be enough?

zoro11031:
> The slow start up issue seems to only happen the first time if there are no suggestions cached. So I seem to hit it a lot on my emulator, but not so much on my actual devices.
> 
> Try adding an initial delay to the `OneTimeWorkRequestBuilder`. Maybe even just 30-60 seconds would be enough?

Gave this a try - it worked but not consistently. If the cache is empty and PeriodicWorkRequest happens to be scheduled to fire around the time you launch the app, you'll still experience start up delays.

I'm thinking maybe just remove `OneTimeWorkRequest` entirely and having `PeriodicWorkRequest` fire immediately with a 30 second delay if the app is started with empty cache. That way we can make sure behavior is consistent.

zoro11031:
@damontecres Give it a try! I launched with empty cache on emulator a few times and it seems like the issue is completely resolved.

damontecres:
@zoro11031 I think there are still issues with performance:
1. Cache has entries, but calling `enqueueUniquePeriodicWork` runs immediately
2. I tried setting the 30s delay always, but if you switch pages while the job is running, the loading for the page will be slow

I think might need to revisit limiting the max number of simultaneous API calls so that loading pages isn't completely blocked. Though thinking more on this, during the API calls, the coroutine should suspend, allowing other `Dispatcher.IO` coroutines to run. So I wonder if something is using CPU time or if we're hitting some JF server max request limit or something.

zoro11031:
> @zoro11031 I think there are still issues with performance:
> 
>     1. Cache has entries, but calling `enqueueUniquePeriodicWork` runs immediately
> 
>     2. I tried setting the 30s delay always, but if you switch pages while the job is running, the loading for the page will be slow
> 
> 
> I think might need to revisit limiting the max number of simultaneous API calls so that loading pages isn't completely blocked. Though thinking more on this, during the API calls, the coroutine should suspend, allowing other `Dispatcher.IO` coroutines to run. So I wonder if something is using CPU time or if we're hitting some JF server max request limit or something.

What if we had it process libraries sequentially instead of all in parallel? With current behavior, the more libraries a user has, the more simultaneous API calls there would be. That adds up fast, especially if they're splitting things up into sub-categories like

- Movies
- 4K Movies
- Anime Movies
- TV Shows
- Anime
etc

PR #842, opened 2026-02-07
Navigation drawer improvements

### LINKED ISSUE #384: [BUG] - When you select something in the side menu, the screen moves and it feels very weird.
### Description

I don’t know how to explain it, but when you open the side menu or simply select an option from it, hiding the menu and returning to the selected screen should be a smooth, continuous effect. In my case, the menu hides in a choppy way and the screen makes strange movements — it doesn’t feel smooth.
I don’t know if it’s just an issue on my side. It would be great if those small details could be refined, as they make the interface feel less polished.

### Reproduction steps

_No response_

### App Version

v0..3.4.9

### Server Version

10.11.4

### Device

NVIDIA SHIELD TV

### Logs

```bash

```

### LINKED ISSUE #699: [UI] Slide content when nav drawer opens
### Discussed in https://github.com/damontecres/Wholphin/discussions/694

<div type='discussions-op-text'>

<sup>Originally posted by **kiwichick888** January 13, 2026</sup>
Is it possible to have the sidebar menu behave like Plex when it's expanded? When expanded in Wholphin, the menu covers some of the current view. When expanded in Plex, the view is 'pushed' to the right so that none of it is covered.
Untitled 1</div>

damontecres:
## Description
This updates the navigation drawer with better behavior and UI. It is also a bit more optimized and is slightly faster.

- When open, content is pushed to the right instead of being covered
- When open, it uses the same background colors from the dynamic background
- The icons no longer shift as the drawer opens & closes
- The open drawer is not as wide
- Focus highlight on items is less circular, now matching the other menus in the app
- The animation is a bit faster too

Dev notes:
- This PR copies & modifies `tv-material` code, so there's a customized `ModalNavigationDrawer` implementation now
- There are fewer `AnimatedVisibility` composables which improves performance
- Icons are exactly sized
- Content pages generally shouldn't add start/end padding anymore, the padding is now added by the nav drawer (b4f1b111edcd7e7ec05edc1100402b2851721fcd) and is consistent across every page

### Related issues
Closes #699
Closes #384

### Testing
Tested on the emulator, nvidia shield, & google sabrina

## Screenshots
nav_drawer_closed Large
nav_drawer_open_backdrop_dim Large

## AI or LLM usage
None

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/nav/ApplicationContent.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/nav/ApplicationContent.kt
@@ -11,13 +11,15 @@ import androidx.compose.runtime.getValue
 import androidx.compose.runtime.saveable.rememberSerializable
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
-import androidx.compose.ui.draw.alpha
 import androidx.compose.ui.draw.drawBehind
 import androidx.compose.ui.draw.drawWithContent
 import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.graphics.BlendMode
 import androidx.compose.ui.graphics.Brush
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.graphics.CompositingStrategy
+import androidx.compose.ui.graphics.SolidColor
+import androidx.compose.ui.graphics.graphicsLayer
 import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.platform.LocalContext
 import androidx.hilt.lifecycle.viewmodel.compose.hiltViewModel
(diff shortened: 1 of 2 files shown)


Pentaphon:
Huge improvement. Only possible issue I see is the gap in the background shown by the red rectangle, which doesn't seem to be what Plex does. @YogiBear12 @zoro11031 any other suggestions from your design-minded eyes?

gap

damontecres:
So Plex doesn't dim the content when the nav drawer is open. The older implementation Wholphin uses that squished the content didn't either.

But the "modal" version that covered part of the content did which makes sense since it was covering the content as if it was a dialog almost. I kept this, but I also like how it looks without the dimming effect.

Also the dimming gradient can be changed, @YogiBear12 is probably better at that than I.

Without dim:
nav_drawer_open_no_dim Large

Pentaphon:
>Without dim:

I much prefer this. Looks way more cohesive. And it looks much more like how Plex does it, according to the pics from the original issue. Thanks!

plex

damontecres:
> > Without dim:
> 
> I much prefer this. Looks way more cohesive. And it looks much more like how Plex does it, according to the pics from the original issue. Thanks!

Yeah, I agree. Been playing around with it for the past 10-15 minutes and I prefer the look without dimming. I'll push this change.

YogiBear12:
Could we dim the backdrop itself? When the navdrawer is open. This is similar to what Plex does and solves the issue of text being hard to read when it's pushed over the backdrop.

Granted, no one should be trying to read the overview when the navdrawer is open, but it's still a small QoL thing. 

Otherwise, I really like this change.

damontecres:
> Could we dim the backdrop itself? When the navdrawer is open

Good idea. Done & updated the image in the PR description.

PR #846, opened 2026-02-08
Hold-to-timeline seek behavior and hold multiplier

### LINKED ISSUE #522: Advance by holding left/right

voc0der:
## Description

Implements "hold-to-timeline" feature where holding left/right directional buttons for 2 seconds shows playback controls and moves focus directly to the timeline seekbar for scrubbing.

**Key changes:**
- Hold detection logic with 2-second threshold in `PlaybackKeyHandler`
- Automatic seekbar focus acquisition when hold triggers
- Event consumption during hold to prevent skip behavior or unwanted events
- Wrap `PlaybackKeyHandler` in `remember()` to preserve state across recompositions
- Ensure no focus theft issue where LaunchedEffect would steal focus back from seekbar
- Synchronous baseline reset for scrub multiplier to ensure scrubbing starts at 1x speed
- Add `SystemNavigationLeft/Right` support to key identifier functions for consistency

### Related issues

Incorporates #784 (seek acceleration)
Related to #522
Closes #522

### Testing

Tested on Nvidia Shield Pro 2019

- Hold left/right directional button for 2s causes the controls to appear, timeline focuses, scrubbing starts at 1x
- Quick tap still performs skip as expected
- Scrub acceleration properly scales with video duration
- No skip behavior occurs during hold

## AI or LLM usage
ChatGPT Codex 5.3, Claude Opus 4.6 used to write much of the code and help debug. All code manually tested and verified.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackPage.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackPage.kt
@@ -225,6 +225,9 @@ fun PlaybackPageContent(
     LaunchedEffect(controllerViewState.controlsVisible) {
         // If controller shows/hides, immediately cancel the skip indicator
         skipIndicatorDuration = 0L
+        if (!controllerViewState.controlsVisible) {
+            seekBarFocusOnShow = false
+        }
     }
     var skipPosition by remember { mutableLongStateOf(0L) }
     val updateSkipIndicator = { delta: Long ->
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackKeyHandler.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackKeyHandler.kt
@@ -148,6 +148,7 @@ class PlaybackKeyHandler(
                 }
             } else {
                 // When controller is visible, its buttons will handle pulsing
+                result = false
             }
         } else if (isMedia(it)) {
             when (it.key) {


app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackPage.kt
```diff
@@ -391,6 +413,11 @@ fun PlaybackPageContent(
                     onPlaybackActionClick = onPlaybackActionClick,
                     onClickPlaybackDialogType = { playbackDialog = it },
                     onSeekBarChange = seekBarState::onValueChange,
+                    seekBarFocusRequester = seekBarFocusRequester,
+                    shouldFocusSeekBar = seekBarFocusOnShow,
+                    onSeekBarFocusConsumed = {
+                        seekBarFocusOnShow = false
```
damontecres: I don't think this should be bubbling this up so far.

The `PlaybackOverlay` composable takes `seekBarInteractionSource` if you need to check for focus.

voc0der:
Good call on the bug. I am able to reproduce it.  I do generally agree with state of it, and how it was done, so definitely not ready for primetime.

I'm going to drop this back into draft state and probably close it while I reconcile the below and probably open a different PR with that in mind.

> I think adding long clicking left/right D-Pad to skip by the repeat count is a good idea. And I'd consider the acceleration idea for both d-pad & seek bar skipping (proposed in https://github.com/damontecres/Wholphin/pull/784 right?).

PR #848, opened 2026-02-08
Fix glitchy nav drawer scrolling when opening

damontecres:
## Description
The nav drawer would always glitch a little when it gained focus due it to trying to keep the selected page in view. #842 made this even worst.

This PR fixes that issue by hoisting the list scroll state up higher which ensures it is remembered between recompositions. This change also eliminates the need for workarounds to try and scroll it programmatically which is also a tiny performance boost.

### Related issues
Follow up to #842 

### Testing
Emulator & nvidia shield

## Screenshots
N/A

## AI or LLM usage
None


damontecres:
This turns out this is a pretty decent performance boost. It reduces the number and length of janky frames pretty significantly on the home page when scrolling.

PR #849, opened 2026-02-08
Fix: optimize suggestions performance

zoro11031:
## Description
- Aims to further optimize performance for the suggestions fetching

### Related issues
Fixes #833 

### Testing
Android Emulator
NVIDIA Shield Pro 2019

## AI or LLM usage
AI being used to assist with updating/creating unit tests

PR #851, opened 2026-02-09
Rewrite SuggestionsCache & limit parallelism

damontecres:
## Description
A rewrite of the `SuggestionsCache` to simplify it as a read-through cache. The downside of this is that updated suggestions won't be reflected in the UI unless the user reloads the page, but I think that's an okay trade off.

Also adds parallelism limits to fetching suggestions to reduce the number of simultaneous API calls. And moved the combining logic to use the `Default` dispatcher to free up an IO one.

Also adds an initial delay to both the suggestions & tv provider workers, so a cold start of the app isn't starved.

### Related issues
Related to #849

### Testing
Emulator & unit testing

## Screenshots
N/A

## AI or LLM usage
None


damontecres:
@zoro11031 What do you think of this? I think we might want to integrate this and #849.

zoro11031:
> @zoro11031 What do you think of this? I think we might want to integrate this and #849.

I think this is a pretty good compromise. Most users won’t even know/care that their suggestions don’t update until a refresh as long as something was already there. Nice work!

PR #900, opened 2026-02-14
Hold-to-seek acceleration for playback DPAD and seekbar

### LINKED ISSUE #522: Advance by holding left/right

voc0der:
## Description
- Adds hold-to-seek acceleration for left/right DPAD input during playback skip handling.
- Reuses one shared, duration-aware acceleration profile for both playback DPAD and seekbar skipping.
- Extracts acceleration logic into a shared helper to keep playback and seekbar behavior consistent.
- Slows the acceleration ramp (about one-third) to improve control during longer holds.
- Preserves tap-on-release behavior: single taps seek on key release, while holds seek on repeated key-down without an extra release step.
- Normalizes unknown/unset duration values to use the shortest-content acceleration profile.
- Added unit coverage for shared seek acceleration math to prevent accidental regression in ramp thresholds.

### Related issues
Supersedes #846  
Incorporates #784  
Closes #522

### Testing
Tested on Nvidia Shield Pro 2019.

- Quick left/right tap still performs normal skip behavior.
- Holding left/right seeks repeatedly with progressive acceleration.
- Seekbar hold uses the same acceleration profile as playback hold.
- Releasing after a hold does not trigger an extra seek step.

## AI or LLM usage
Codex 5.3 and Claude Opus 4.6 were used for implementation, debugging, and cross-review. Changes were manually validated on device.


app/src/main/java/com/github/damontecres/wholphin/ui/playback/SeekAcceleration.kt
```diff
@@ -0,0 +1,51 @@
...
+
+/**
+ * Shared seek acceleration profile for hold-to-seek behavior.
+ * Keep this in sync anywhere directional key repeat seeking is handled.
+ */
+fun calculateSeekAccelerationMultiplier(
+    repeatCount: Int,
+    durationMs: Long,
+): Int {
+    if (repeatCount <= 0) return 1
```
damontecres: Fail fast if duration is not valid

```suggestion
    if (repeatCount <= 0 || durationMs < 0) return 1
```

damontecres (commented):
Thanks for the contribution, this is solid! I left a few minor suggestions.

I think the ramp up multiplier for <90m could start a little sooner. On my shield with 30s skip forward & ~41m show, the multiplier didn't start until around minute 38, about 2/3 of the show. I think maybe starting slightly sooner at 40-50% mark? I think the options for >90m are okay.

It also could be interesting if the hold-to-skip started slower. Like maybe wait until the repeat count is 10-15? Though I'd have to try it to see if it feels natural so up to you if you want to try implementing that or not.

Thanks again!

voc0der:
Adjusted the <90m ramp to start earlier (first threshold 25 -> 13) so x2 kicks in closer to the 40-50% mark on shorter episodes. I left the >90m tiers unchanged.

Also added a small no-op window before hold-repeat seeking starts so hold-to-skip comes in more gradually, and made sure the acceleration ramp doesn’t advance during that no-op window.

Going to continue testing this and improving it. Thanks for the feedback!

voc0der:
I tested this current code and it seems to perform quite well. I won't mark anything resolved since I want to hear what your thoughts are.

Unknown duration are now excluded since the more I reasoned with it, the less it made sense, after seeing your review.

Edit: I will also test this without the composition wait soon.

Edit2: Testing the SeekBar without composition, the regular seek worked identically, you were right to call that out.

Edit3: Both features seem to work the same way and with no observable flaws.

Pentaphon:
I have added some suggestions for improvement to this feature here: https://github.com/damontecres/Wholphin/discussions/1122

Issue #528, opened 2025-12-21
[FEA] - Video Player UI Modernization (Long-term goal)


Pentaphon:
### Description

Right now, the biggest weak point in terms of UI is Wholphin's video player. It works and looks fine, but it's not a standout in terms of UI. Admittedly, this is a big feature request so please consider this a long-term goal. 

How can Wholphin's video player feel more modern, polished, and intuitive? More like a commercial streaming service player than what we currently have, which is a bit on the basic side. What are realistic changes that can be made to make the player nicer looking and more intuitive than any other Jellyfin client? 

My thoughts for player improvements:

1. Themed player buttons of different sizes and shapes, possibly rounded, modern icons. 
2. When scrubbing, a scrub-to-time display that doesn't overlap with the episode info. **EDIT FIXED**
3. Smaller, more compact menus for audio, playback speed, video scale, sub delay that don't take as much space on the screen yet are still very readable.
4. Episode and movie info under the title logo on the top left of the player UI
5. Putting the clock and end time together at the top right corner. 
6. A more compact, yet readable debug info display. 
7. Grouping the menu buttons together, away from the player buttons.
8. More use of the theme color throughout the UI like the player menus, debug info, etc. 
9. the "3 dots button" on the left only brings up a single menu item (the debug info) instead of being an "I" for info that opens the info display and turns into the theme color when activated. 

I'm hoping @YogiBear12 has some ideas but I'm leaving this open to anybody reading this issue and hopefully with all our input, we can flesh out a better video player, which can then be translated into a similar-looking music player which is coming according to the roadmap. 

Thoughts?

damontecres:
I generally like the playback controls as they are. I made it this way because I like it. It's similar to Plex (shocker).

I always am open to ideas and suggestions, but I'll give some guidelines for this.

## Things that need improvement:
- The menus, or "bottom dialogs" as I called them in the code, are large & clunky and not keeping with the color themes
    - The highlighted item's borders don't match the overall dialog's borders
    - The selected item (subtitle) has an almost imperceptible color change in addition to the circle indicator
- Reorganizing the buttons such as:
    - Move subtitle button to left side & add an audio button to select audio
    - Combine the left "more" button with the settings button, so it would have a popup for debug info, playback speed, & scale settings

## Things I do not like:
- Don't group all of the buttons to one side (like the official app), I prefer a more symmetrical look
- Don't use icon-only buttons that are not universally recognizable
[...]

PR #923, opened 2026-02-19
Fixing seek text label overlap and visibility issues.

crisc98:
## Description
- Smoothly bumps up the title and subtitle text above the seek progress label when the seek bar is focused to prevent overlap.
- Adds a semi-transparent black backer to the text label that appears above the seek progress indicator. Helps with visibility when the image in the playback matches the color of the text. Note: the backer is black with Alpha of 0.6 for now because I felt that looked good for contrast across all themes. However, if you prefer it respect a particular theme color, I'd be willing to make that change.

### Related issues
In the playback screen, when focusing the seek bar on Android TVs (at least the ones I was able to test), the seek text above the seek bar will overlap with the media title. This is a purely cosmetic issue in the UI, but it seemed worth a fix. See before and after images in in the Screenshots section below

This problem is briefly mentioned in point number 2 of [#528 ](https://github.com/damontecres/Wholphin/issues/528)

### Testing
Tested manually on TCL model 50S434 running Android TV. Tested manually using AVD emulation on Television (1080p) and Television (4K) preset devices. Reproduced the problem on all 3 devices and confirmed that my fix works across all.

## Screenshots
Before:
before

After:
after

## AI or LLM usage
None

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackOverlay.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackOverlay.kt
@@ -449,7 +449,7 @@ fun PlaybackOverlay(
                         modifier =
                             Modifier
                                 .background(Color.Black.copy(alpha = 0.6f), shape = RoundedCornerShape(4.dp))
-                                .padding(horizontal = 6.dp, vertical = 2.dp),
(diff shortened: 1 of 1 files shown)


Pentaphon:
Thanks for this. A much-needed, yet simple improvement. One of many little things that can be done to really bring the player to a whole new level.

PR #928, opened 2026-02-19
Fixing Settings sticky header text overlapping with menu items

crisc98:
## Description
- Adding a background for the Settings stickyHeader. This helps with visibility and prevents overlap with the menu items when scrolling.
### Related issues
In the Settings screen, the header can overlap the text of the menu items when scrolling down. This is a minor UI bug only and does not affect functionality at all, but the fix seemed easy and harmless so I figured I'd give it a go.

### Testing
Tested manually on TCL model 50S434 running Android TV. Tested manually using AVD emulation on Television (1080p) and Television (4K) preset devices. Reproduced the problem on all 3 devices and confirmed that my fix works across all.

## Screenshots
Before:
before-settings-fix
After:
after-settings-fix

## AI or LLM usage
None

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/preferences/PreferencesContent.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/preferences/PreferencesContent.kt
@@ -11,8 +11,10 @@ import androidx.compose.foundation.interaction.MutableInteractionSource
 import androidx.compose.foundation.interaction.collectIsFocusedAsState
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
+import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.PaddingValues
 import androidx.compose.foundation.layout.fillMaxHeight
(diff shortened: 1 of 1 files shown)


damontecres:
Thanks for the contribution! I just never got around to fixing this.

However, I think a better way would be to remove the `stickyHeader` and wrap the `LazyColumn` in a regular `Column` and put the header text in the `Column` above the `LazyColumn`.

That's what I've started doing in [newer parts of the app](https://github.com/damontecres/Wholphin/blob/main/app/src/main/java/com/github/damontecres/wholphin/ui/main/settings/HomeRowSettings.kt#L42-L44).

crisc98:
@damontecres Thanks for the feedback! I applied your suggestions in my latest commit. Diff is a little wild because of the hierarchy changes, but it seems to work well and looks even better than my original fix!

damontecres:
> Diff is a little wild because of the hierarchy changes

On the Files changed tab, you can click the gear icon at the upper right side and enable "Hide whitspace" to make changes like this easier to view! I think the setting is per PR.

Thanks again!

PR #946, opened 2026-02-21
Add in-app & Android TV screensaver

### LINKED ISSUE #230: [FEA] - Screensaver when idle
### Description

Like the official android tv client, it would be nice to have a screensaver kick in after some idle time (instead of always showing the paused screen)
- The screensaver would contain movie/show title, background image, and clock. The title can just be text or the logo (like the official android tv client)

Additional features:
- If screensaver started while on the home screen, clicking the d-pad / ok button would take you to the media
- Option to only show unwatched movies (and series with unwatched episodes)

damontecres:
## Description
Adds both an optional in-app and Android TV OS (DreamService) screensaver

Has settings for:
- Starting delay, options from 30s to 1 hour
- Duration to show each image, 30s to 2 minutes
- Show clock (separate from app-wide clock)
- Toggle animated zooming
- Set a max age rating to limit what content is shown
- Included media types: movies, tv shows, and/or photos

By default only Movies & TV Shows are included. But if you have like family albums on your server, you could enable Photos too.

There's also an button to start the screensaver so you can see what it looks like.

### Related issues
Closes #230

### Testing
Emulator, Onn 4k pro, nvidia shield using both in-app and OS screensavers

## Screenshots
screensaver_settings Large

## AI or LLM usage
None

## Acknowledgements

Some of the skeleton code for setting up the `DreamService` with the right lifecycle was copied and modified from the [official app's implementation](https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/java/org/jellyfin/androidtv/integration/dream/DreamServiceCompat.kt).

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/WholphinDreamService.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/WholphinDreamService.kt
@@ -74,7 +74,7 @@ class WholphinDreamService :
                             val currentItem by itemFlow.collectAsState(null)
                             AppScreensaverContent(
                                 currentItem = currentItem,
-                                showClock = prefs.appPreferences.interfacePreferences.showClock,
+                                showClock = screensaverPrefs.showClock,
                                 duration = screensaverPrefs.duration.milliseconds,
                                 animate = screensaverPrefs.animate,
                                 modifier = Modifier.fillMaxSize(),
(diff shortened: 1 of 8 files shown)


Hukuma1:
Just wanted to say a big thank you for this. Working beautifully!

PR #1014, opened 2026-02-27
Add option to delete media

### LINKED ISSUE #104: [FEA] - Option to delete media
### Description

In the android official jellyfin mobile app there is an option within each movies info page to "delete media". It provides a quick way to delete specific movies and actually deletes from the server device. Would be appreciated if this was added to wholphin

damontecres:
## Description
Adds a setting to enable "Media management" which adds a context menu item to delete media. This is accessible using either the More button or long clicking.

You can delete episodes, seasons, series and movies/videos.

If order for the option to appear the current user must have server-side settings for "Allow media deletion from" enabled for the library and the in-app setting under Settings->Advanced Settings->"Show media management options".

### Related issues
Closes #104 

### Testing
Emulator

### Context menu
delete_context_2 Large

### Confirmation
delete_confirm Large

## AI or LLM usage
None

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/components/Dialogs.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/components/Dialogs.kt
@@ -19,11 +19,14 @@ import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
+import androidx.compose.foundation.layout.wrapContentWidth
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.LazyListScope
 import androidx.compose.foundation.lazy.itemsIndexed
 import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.foundation.shape.RoundedCornerShape
+import androidx.compose.material.icons.Icons
+import androidx.compose.material.icons.filled.Delete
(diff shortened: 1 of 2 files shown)


Pentaphon:
Thanks for this. Only issue I see is that the delete confirmation prompt is a bit repetitive. Instead of the word "delete" in the header followed by the line of text asking to confirm you could replace the header entirely with just the red trash icon like so. I could be wrong but just a suggestion to cut down on text and use more icons when possible. 

del

damontecres:
I updated the dialog to emphasize which media item more strongly and put the trash icon in the confirm button.

image

Pentaphon:
> I updated the dialog to emphasize which media item more strongly and put the trash icon in the confirm button.

Even better. I wasn't sure how you felt about icons in buttons.

JumboMite:
@damontecres new delete option is working, but a couple of things. 

1) the delete button appears on the main TV episode screen (and context menu) but it's missing from the main movie screen (is present in the context menu).

2) the delete option is missing from the context menu on home for both TV shows and Movies.

3) On TV shows tab library, delete option is missing from context menu for TV Shows. It is present for TV show season and TV show episode here.

4) Similarly for Movies tab library, delete is missing from the context menu.  I first need to click on the movie to get to the movie details screen, tap on the 3 dots to bring up the context menu to reach delete.

I can make a screenshot of any of that isn't clear.

damontecres:
> new delete option is working, but a couple of things.

Thanks, #1040 addresses these and then some

Issue #1049, opened 2026-03-06
ASS/SSA subtitles not rendering in direct playback mode after libass-android integration (poor rendering in fallback mode)


Codling:
### Description

Following the integration of libass-android for proper SSA/ASS subtitle support, ASS/SSA subtitles no longer appear at all when using direct rendering / hardware overlay / ExoPlayer direct playback mode.
In the fallback rendering mode (software rendering / surface view / non-direct), subtitles are visible but render poorly — similar to the pre-libass behavior (basic text without full styling, missing karaoke effects, incorrect positioning, no animations, etc.).

### Media info

General
Unique ID                                : 321146939695240394941420091030331557474 (0xF19AA698E93F30506122C89571118662)
Complete name                            : C:\[Moozzi2] Kusuriya no Hitorigoto S2 - 06 [ 30 ] (BD 1920x1080 x265-10Bit Flac).mkv
Format                                   : Matroska
Format version                           : Version 4
File size                                : 822 MiB
Duration                                 : 24 min 1 s
Overall bit rate mode                    : Variable
Overall bit rate                         : 4 785 kb/s
Frame rate                               : 23.976 FPS
Encoded date                             : 2026-03-06 15:37:43 UTC
Writing application                      : mkvmerge v97.0 ('You Don't Have A Clue') 64-bit
Writing library                          : libebml v1.4.5 + libmatroska v1.7.1

Video
ID                                       : 1
Format                                   : HEVC
Format/Info                              : High Efficiency Video Coding
Format profile                           : Main 10@L4@High
Codec ID                                 : V_MPEGH/ISO/HEVC
Duration                                 : 24 min 1 s
Bit rate                                 : 3 517 kb/s
Width                                    : 1 920 pixels
Height                                   : 1 080 pixels
Display aspect ratio                     : 16:9
Frame rate mode                          : Constant
Frame rate                               : 23.976 (24000/1001) FPS
Color space                              : YUV
Chroma subsampling                       : 4:2:0
Bit depth                                : 10 bits
Bits/(Pixel*Frame)                       : 0.071
Stream size                              : 605 MiB (74%)
Writing library                          : x265 3.6+13-0ed28893d:[Windows][GCC 14.1.0][64 bit] 10bit
[...]

PR #1052, opened 2026-03-06
Integrate with libass-android to support SSA/ASS subtitles in ExoPlayer

### LINKED ISSUE #1107: ASS subtitles not rendering correct
### Description

II'm not sure if this issue might already be resolved in #1048. So sorry if this is a duplicate.

ASS subtitles appear as plain black text when used with Exoplayer, and sometimes there are rendering errors (characters and lines that shouldn't be there). If I disable libass, the subtitles display correctly. There are no issues with MPV.

### Reproduction steps

1. Use libass for ASS subtitles on ExoPlayer
2. Media has ASS subtitles
3. Broken subtitles

### App Version

v0.5.3

### Server Version

10.11.6

### Device

Google TV Streamer

### Logs

```bash

```

damontecres:
## Description
Add better ASS/SSA support to ExoPlayer by using [`libass-android`](https://github.com/peerless2012/libass-android).

This is enabled with the "Use libass for ASS subtitles" advanced settings for ExoPlayer. It is enabled by default.

### Related issues
Related to #22
Replaces reverted #569
Fixes #1107

### Try it out

See [instructions here](https://github.com/damontecres/Wholphin/releases/tag/develop-libass)


ethaldeman:
@damontecres I have tested the pre-release of this build and for me it is 100% working. I tested it with various tv shows and my test mkv which includes complex karaoke animations and utilizes embedded fonts. 

While I hope others can test and report back to support my findings, I believe it is ready to be merged.

nx6:
This build does utilize fonts, including fonts attached to mkvs with application/octet-stream instead of more "correct" methods. A clear improvement over ExoPlayer's own support before. 👍 

Performance is comparable to other players utilizing ASS with ExoPlayer. For most tasks it is fine, but the slight delay in rendering it exhibits will cause issues for signs. Particularly more complicated ones with masking where frame-exact timing is needed. For simpler work it's very serviceable. Again, this is an issue on all players that implement this ASS support in ExoPlayer. Mpv will try to maintain rendering to timing accurately despite limited performance of streaming devices, usually at the expense of smooth video playback.

Examples: https://files.catbox.moe/ja9sdz.7z

- Ex1 - Signs on stalls and banners in background at a festival (masking around characters and timed to end on scene change).
- Ex2 - Translation of smartphone screen motion tracked, and timed to the zooms in the latter half.
- Ex3 - Simple text matched to an object in motion (text on a book cover).
- Ex4 - More extreme on-screen text motion tracking/masking.

damontecres:
Thanks @ethaldeman & @nx6, I appreciate the testing feedback for ExoPlayer!

Have either of you tried using MPV on this build as well? The MPV player will use this `libass` build as well which technically shouldn't be significantly different, but you never know.

And no one has run into anything like #1049? This bug report is why I reverted this change originally.

nx6:
> And no one has run into anything like #1049? This bug report is why I reverted this change originally.

Have not seen anything like that.

> Device
> Sony Bravia 9

Kinda leaning towards that being [a device issue](https://forums.plex.tv/t/ass-subtitles-on-android-tv-sony-doesnt-render-when-italic/923718), honestly. Would be good if they could compare to other apps (and supplied a sample). I've only tested this build on one device myself though (a Fire Stick 4K Max, Gen 2) so possible something will happen somewhere else, but I doubt it. Can check more later today.

ethaldeman:
It seems my "100% working" may have been a little premature.......

> And no one has run into anything like https://github.com/damontecres/Wholphin/issues/1049? This bug report is why I reverted this change originally.

I too have not seen anything like that.

https://www.dropbox.com/scl/fi/uss9qa7hxzskoguodc0sn/ASS_Sample_Test.mkv?rlkey=038nzj2czoxnj5vymmth54vrn&st=o9lx2o7n&dl=0

Above is a link to one of my test mkv that is  copyright free and free use. It tests various karaoke effects, but is not a good test of frame perfect complex signs.

I downloaded nx6's 4 sample mkv and tested them. All four had the ASS Sign's render correctly, but they were pretty off timing wise, roughly .5 seconds. I also tested the official jellyfin android tv implementation of libass-android and it had the same issues. I also tested in Wholphin's MPV and official jellyfin android tv ffmpeg transcoding. Both of these did not have the timing issues. Thus it seems to be an libass-android Exoplayer issue. All my testing was performed on the Onn 4k Pro (Amlogic S905X5M and 2 GB RAM)

I plan on opening an issue upstream in libass-android, noting this may not be fixable in Exoplayer citing https://github.com/androidx/media/issues/2289. However in the meantime, I still think the functionality is good enough in most use cases to be merged, maybe just not enabled by default, i.e the default is Exoplayer will still transcode ASS.

nx6:
> I plan on opening an issue upstream in libass-android, noting this may not be fixable in Exoplayer citing [androidx/media#2289](https://github.com/androidx/media/issues/2289). However in the meantime, I still think the functionality is good enough in most use cases to be merged, maybe just not enabled by default, i.e the default is Exoplayer will still transcode ASS.

This has been a known issue for awhile. The Plex dev primarily responsible for the ASS support on the client there attributed it to limitations of ExoPlayer itself, essentially saying that it wasn't capable of frame-exact timing on subtitles. They wanted to work on improving the efficiency in rendering ASS subs but that never came to pass (this was a few years ago).

nx6:
> I've only tested this build on one device myself though (a Fire Stick 4K Max, Gen 2) so possible something will happen somewhere else, but I doubt it. Can check more later today.

I've had the opportunity to try this on a _first_-gen Fire Stick 4K Max and a Shield Pro now. I was going to report that on the Shield two of my four test clips are not using the included fonts, but after some more tests and looking at my Dashboard I see now that those two clips are not direct playing and the Jellyfin server is incorrectly rendering the subs in the transcode. Ex3 and Ex4 are Hi10p video (10-bit h264), so I should have known better since the Shield cannot direct-play that format.

I have not messed with the mpv setting on the first-gen Max, because there is a known chroma-corruption issue on 10-bit HEVC, which is the most-common type so really reduces the usefulness. This is not a Wholphin issue, it's device-specific and impacts Plezy and mpv-android, too (but no such issue on the second-gen Max). The Shield with mpv setting played back subs as expected (better performance than any sticks but not up to a normal PC's performance).

The only thing I've really noticed is alpha-value rendering seems to be slightly off. In my Ex1 there's masking over the names of the stalls in the background ("Octo-Balls" and "Candy Apples") the ExoPlayer render is allowing the original Japanese banner to show faintly throgh. In mpv the masking appears fully opaque. But this is very minor and doesn't impact readability.

nx6:
> The only thing I've really noticed is alpha-value rendering seems to be slightly off. In my Ex1 there's masking over the names of the stalls in the background ("Octo-Balls" and "Candy Apples") the ExoPlayer render is allowing the original Japanese banner to show faintly throgh. In mpv the masking appears fully opaque. But this is very minor and doesn't impact readability.

Ehhh, maybe I will file a bug against libass-android for this...

edge

announcement

This is outside the performance level of most streaming devices. Nonetheless, clip here: [sample.zip](https://github.com/user-attachments/files/26692710/sample.zip)

Edit: Filed as [issue 74](https://github.com/peerless2012/libass-android/issues/74) there.

damontecres:
I appreciate everyone's feedback and help with testing this!

Based on my testing and your feedback, I feel like using libass is not 100% perfect, but it is a big improvement over the default ASS support in most cases. 

I do plan to merge this soon. Right now I have the settings so using libass is the default, but this can be changed back to ExoPlayer's built-in (previous default) or server-side transcoding.

PR #1111, opened 2026-03-17
Allow switching of ffmpeg module tags

brianjmurrell:
## Description
Since the clones of the modules for the ffmpeg build are for a single tag only, allow checkouts to switch to new tags.

### Related issues
https://github.com/damontecres/Wholphin/pull/1106#issuecomment-4077678690

### Testing
Tested on my local build of Wholphin on Linux switching the versions of the various modules.


damontecres:
Check if #1110 works well enough for you.

I guess this change actually creates the local ref instead of using a detached head though.

brianjmurrell:
> Check if #1110 works well enough for you.

I guess I can't until there is another version bump since I already have the updated `media` tag.

> I guess this change actually creates the local ref instead of using a detached head though.

There is that, for whatever value it provides.

damontecres:

PR #1130, opened 2026-03-22
Better playback overlay appearance animations

damontecres:
## Description
Updates the playback overlay so that individual component animate in more deliberately.

For example, the logo & clock slide in from the top instead of the middle of the screen. And the trick play preview now expands and shrinks vertically.

### Related issues
I thought there was an issue about this, but I can't find it

### Testing
Emulator, nvidia shield

## Screenshots
N/A

## AI or LLM usage
None


Pentaphon:
> I thought there was an issue about this, but I can't find it

It was mentioned in a comment or two on different issues that I can't find, either. The "player should slide into place in 2 parts from the top and bottom instead of all at once from the bottom" has been mentioned the most. I was trying to keep all video player progress talk here https://github.com/damontecres/Wholphin/issues/528 but it hasn't taken off as much as I wanted it to. Feel free to convert it to a discussion.

Issue #877, opened 2026-02-11
[BUG] - UI Issue


husoyo:
### Description

In libraries, when details is enabled, bottom of the posters are filling the top, creating a bad UX. 

![Image](https://github.com/user-attachments/assets/9287fdeb-bf3b-4729-a1ae-ce72f2c5e267)

### App Version

v0.4.2

### Server Version

10.11.6

### Logs

```bash

```

damontecres:
Possibly the fix in #882 could be adapted to fix this, but I need to investigate how it interacts with grid scrolling.

husoyo:
> Possibly the fix in #882 could be adapted to fix this, but I need to investigate how it interacts with grid scrolling.

One more note: I realized you can change the view for each collection seperately but there is no one option to change them all at the same time. That'd be great to have.

Issue #737, opened 2026-01-21
[FEA] - Advanced BoxSet/Collection View


kamilkosek:
### Description

When you open a BoxSet/Collection it displays the items in the default `CollectionFolderGrid` making it quite difficult to spot new releases, recently added content, suggestions etc..

<img width="917" height="578" alt="Image" src="https://github.com/user-attachments/assets/4aa5a89a-b7a3-45d0-a706-9f21efa50476" />

Ideally there would be an option to enable `Advanced View Mode for BoxSet/Collections` so that these will be displayed more like libraries:

<img width="915" height="552" alt="Image" src="https://github.com/user-attachments/assets/10049133-7ed6-410f-802b-12ad8fb79b7f" />

sysmoon14:
Yes! And let me put it on the home screen like Plex

kamilkosek:
@sysmoon14  you mean add a BoxSet/Collection to the home screen as a row ? 
something like this: 

<img width="970" height="549" alt="Image" src="https://github.com/user-attachments/assets/c2d90a44-7ada-4d64-ad19-1522d2d1ed12" />

sysmoon14:
Yep!

damontecres:
Related to #527

Issue #383, opened 2025-12-06
[FEA] - If title logos are enabled, show title logos on all app menus and not just on Player UI


Pentaphon:
### Description

I realize that some people might want text over the logos, but if title logos are enabled on the Player UI, they should also be enabled on the app interface menus as well. In other words, the title logo option should work app-wide. They give the menus a nice splash of color.

PR #1137, opened 2026-03-24
Better collection pages including more view options

### LINKED ISSUE #527: [FEA] - Collection Details (Images & Descriptions)
### Description

If available, Id like to have collection details viewable on the collection page or within the individual collection. Things like Images, overview, and other metadata if available. 

Image

damontecres:
## Description
This is a large update to the collection page!

- Show the collection's details including overview & backdrop
- Separate collection by type (Movies, TV, etc) into rows
- View options are now shared across all collections
- Filter & sort are still saved per collection

By default, the collection will be separated by type, but there is a view option to restore the combined grid.

### Related issues
Closes #527 
Closes #1088
Fixes https://github.com/damontecres/Wholphin/issues/877#issuecomment-3889013124
Related to #737

### Testing
Emulator mostly

## Screenshots
collection_page Large

### Separated types into rows
collection_rows Large

### Combined type into grid
collection_grid Large

## AI or LLM usage
None


Pentaphon:
The only remaining issue I see here is the awkward looking title at the top that currently says "Marvel" in your screenshots. Something about it feels "off" and out of place but I can't figure out a better way to do it.

Hukuma1:
Would be awesome if it said Marvel collection via custom logo. Like here with Thor collection:

image

damontecres:
> The only remaining issue I see here is the awkward looking title at the top that currently says "Marvel" in your screenshots. Something about it feels "off" and out of place but I can't figure out a better way to do it.

I agree. I think I have some ideas though.

> Would be awesome if it said Marvel collection via custom logo

I still struggle with incorporating logos because they can vary in size so much. But I don't disagree, it would be nice.

I plan to follow up this PR later to take another step towards making Collections feel more than just a list such as how Movies are shown.

damontecres:
@Pentaphon @Hukuma1 

I decided to just go for a full rewrite and test out using a logo. It still needs lots of work in terms of exact sizes, padding, and animations. But, what do you think so far?

Basically, when opening a collection, it shows the collection's details up top and uses the collection's backdrop. Then when moving down to the collection's content, the header will switch to the focused items.

Also, I haven't put this up on a TV yet, so sizes might be off.

## Edit: this is outdated!
[collection_example.webm](https://github.com/user-attachments/assets/c204e4ba-55c2-4ddc-a47e-d4c829360f99)

This layout also means I can add buttons/context menu for favoriting, deleting, etc. So that's a good advantage too.

Pentaphon:
> But, what do you think so far?

Looks much better than before. This is closer to what I would have done but I wasn't sure if you wanted to make sweeping changes but you've since done a full re-write. However, I do predict this new UI will prompt users to ask why logos are used for collections and not single movie and TV show titles. You may want to address https://github.com/damontecres/Wholphin/issues/383 with the new code and adapt the logos to all pages in one fell swoop?

Hukuma1:
I think it looks great. But yeah, logos on movie pages and TV shows would be nice too.

I know your issue is with sizing. Perhaps if it's too tiny, don't display one and fall back on text. If it's too large, scale it down? Or would that be too heavy on resources? I don't want to bloat Wholphin either.

damontecres:
I decided to revert the logo change in this PR. I just want this one to focus on a new collection page.

But since I was happy with how the logo looked, I plan to add it back throughout the app in a new PR.

PR #1140, opened 2026-03-24
Add option to remove item from Continue Watching and Next Up rows

### LINKED ISSUE #906: [FEA] - Long press, Remove from continue watching
### Description

How it is right now

1. Go to show
2. Go to series
3. Mark Watched
4. Mark Unwatched
5. It goes off continue watching / next up

How I would like it to be 

1. Long press on title 
2. Remove from continue watching

damontecres:
## Description
Adds context menu options for "Remove from continue watching" and "Remove series from next up"

### Remove from continue watching

This marks the item as unplayed so it will not be shown on the Continue Watching row anymore. It may still show on a Next Up or combined row though.

### Remove series from next up

This first does the same as "Remove from continue watching" above. But it also stores the series' ID in your display preferences as removed. Wholphin will filter out this series from appearing in the Next Up row until you watch _any_ episode in the series in the future.

This only works for Wholphin clients!

There's a dialog to view which series have been removed and to re-add them manually. Settings->Customize Home->Settings->View removed next up

### Related issues
Closes #906

### Testing
Mostly emulator testing

## Screenshots
image

## AI or LLM usage
None


damontecres:
I put this as draft because while it works, I'm not sure I'm fully in favor of how the "Remove series from next up" is implemented especially since it only works for Wholphin. I think there are some plugins like KefinTweaks maybe that support this, so maybe it would be better to integrate somehow.

I'm curious if anyone has any thoughts!

kiwichick888:
It's fantastic you're working on this. My two cents worth:
1. I'd rather remove from continue watching did not mark the item as unplayed. 
2. Interested to know under what circumstances the removed item _may_ still show on a combined row. I wouldn't want that to be the case.

Motophan:
This is exactly what I wanted. 100

Re kikichick888, they have to do it this way because of how jellyfin handles it. There isnt a real option under the hood for remove from continue watching, it just runs two commands (mark as watched, mark as unwatched) because in the actual jellyfin backend there isnt one. Once there is they can just use that component.

kiwichick888:
> they have to do it this way because of how jellyfin handles it. There isnt a real option under the hood for remove from continue watching, it just runs two commands (mark as watched, mark as unwatched)

Oh that's a shame. I hoped it would have the same behaviour as Plex. But, if it can't be done, it can't be done 🙂

PR #1144, opened 2026-03-25
Don't show full playback overlay when pausing

damontecres:
## Description
Instead of displaying the full playback overlay when pausing via a button press, just show a short pause image in the center of the screen. Any button press that will pause triggers this including a remote Pause, Pause/Play, or one-click pause Enter.

I decided to make this the standard behavior instead of a toggle because I think if the user intends to do something with the overlay (change subtitles, audio, etc), it's still straightforward to open the overlay, click the pause button (which is focused first), and then move to the desired option versus assuming click-to-pause will open the overlay so the user can move to the desired option.

### Related issues
Closes #990

### Testing
Emulator

## Screenshots
[pause_osd.webm](https://github.com/user-attachments/assets/8f908164-2a50-43f9-9e13-a463d2914e42)

## AI or LLM usage
None

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/playback/MediaSessionPlayer.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/playback/MediaSessionPlayer.kt
@@ -1,23 +1,23 @@
 package com.github.damontecres.wholphin.ui.playback
 
+import androidx.annotation.OptIn
 import androidx.media3.common.ForwardingSimpleBasePlayer
 import androidx.media3.common.Player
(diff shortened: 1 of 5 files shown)


Pentaphon:
Thank you! This really bridges the gap in the player UI where we had one-tap right and left simple progress icons but no one-tap simple pause icon. The only missing issue I can see is that you may want to have a play button appear in a similar manner when resuming the video to make things consistent in terms of showing a UI element when the user interacts with the video.

brh150:
Thank you! I'm so glad it's got implemented.

PR #1151, opened 2026-03-28
Add prefer logos instead of text titles setting

### LINKED ISSUE #383: [FEA] - If title logos are enabled, show title logos on all app menus and not just on Player UI

damontecres:
## Description
Adds an advanced settings to prefer logos for titles instead of text. It is enabled by default.

### Related issues
Closes #383

### Testing
Emulator & onn

## Screenshots
<details><summary>Click to show</summary>
<p>

### Home page
logo_1 Large

### Series overview
logo_2 Large

### Movie Page
logo_movie Large

### Square-ish logo
logo_square Large

</p>
</details> 

## AI or LLM usage
None

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -747,6 +747,6 @@
     <string name="continue_string">Continue</string>
     <string name="select_all">Select all</string>
     <string name="separate_types">Separate types</string>
-    <string name="prefer_logos">Prefer showing title logos</string>
+    <string name="prefer_logos">Prefer showing logos for titles</string>
 
 </resources>


damontecres:
The logos have a fixed height and go up to ~5x the height in width.

I do think the wider logos could be shorter, but then the ones closer to 1:1 will be too small. I want a fixed height though so that that header doesn't jump around when scrolling between items.

Pentaphon:
Thanks for this! The squarish logos are a bit on the small side. Perhaps logos with a square aspect can have a different fixed height than wide logos so you can have both shorter wide logos and bigger square logos?

damontecres:
> Perhaps logos with a square aspect can have a different fixed height than wide logos

This would cause the header and/or rows to move up and down as you scroll which very distracting.

Maybe I could increased the fixed logo height, but wider images only occupy part of it with padding above/below. One problem is that the API doesn't provide aspect ratios for logos, so the dynamic sizing would be a bit more complicated.

Pentaphon:
> One problem is that the API doesn't provide aspect ratios for logos, so the dynamic sizing would be a bit more complicated.

Ah, then I wouldn't worry about it until the API improves. Thanks again.

PR #1154, opened 2026-03-28
Fix server discovery reporting localhost instead of real IP

baykatre:
When discovering servers on the local network, the app shows localhost or 127.0.0.1 because the Jellyfin SDK reads the Address field from the server's UDP response. Many Jellyfin setups report their loopback address there.

I added a LocalServerDiscovery class that does the same UDP broadcast but uses the actual source IP from the incoming packet. Updated SwitchServerViewModel to use it.

Tested on a network where the server was reporting http://localhost:8096 — now correctly shows http://192.168.0.5:8096.

AI/LLM disclosure: Claude was used to assist with this code change.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/services/LocalServerDiscovery.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/services/LocalServerDiscovery.kt
@@ -1,0 +1,97 @@
+package com.github.damontecres.wholphin.services
+
+import kotlinx.coroutines.flow.Flow
+import kotlinx.coroutines.flow.flow
+import kotlinx.coroutines.flow.flowOn
+import org.json.JSONObject
+import timber.log.Timber
+import java.net.DatagramPacket
+import java.net.DatagramSocket
+import java.net.InetAddress
+import java.net.SocketTimeoutException
+import java.net.URL
+import java.util.UUID
+import kotlin.coroutines.CoroutineContext
+
+/**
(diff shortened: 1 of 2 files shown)


damontecres:
Thanks for the suggestion. After review, this is a rewrite of the SDK's [`LocalServerDiscovery`](https://github.com/jellyfin/jellyfin-sdk-kotlin/blob/master/jellyfin-core/src/jvmCommonMain/kotlin/org/jellyfin/sdk/discovery/LocalServerDiscovery.kt).

I don't want Wholphin to have to maintain code like this. You should you raise this as an issue or open a PR in the SDK repo itself instead: https://github.com/jellyfin/jellyfin-sdk-kotlin

baykatre:
Thanks for the review. You're right — I've opened a PR on the SDK itself instead: https://github.com/jellyfin/jellyfin-sdk-kotlin/pull/1209

The Jellyfin server doesn't populate EndpointAddress in its UDP discovery response — it only sends Address, Id, and Name (see https://github.com/jellyfin/jellyfin/issues/11564). Other client SDKs (Python, WebOS) fill endpointAddress from the UDP packet source IP on the client side, but the Kotlin SDK never does. Other clients like Delfin are also implementing their own UDP discovery to work around this (https://codeberg.org/avery42/delfin/pulls/137). The SDK PR is a single-line change to fix that.

If the SDK PR isn't accepted, Wholphin can still work around this minimally by just overriding endpointAddress on the ServerDiscoveryInfo objects returned from the SDK's discovery flow — no need to reimplement the full discovery logic.

PR #1164, opened 2026-03-31
Quality: JNI method cache initialization is not thread-safe and can publish partially initialized IDs

### LINKED ISSUE #1163: fix: jni method cache initialization is not thread-safe and can publish partially initialized ids
## Description

`init_methods_cache` uses a plain static boolean (`methods_initialized`) as a guard without synchronization. If two threads enter concurrently, one thread can observe `methods_initialized` before all global class/method IDs are fully assigned, causing later JNI calls to use null or stale `jclass`/`jmethodID` values. In native code this can lead to hard crashes (SIGSEGV/abort) during `CallStatic*Method`.

**Severity**: `high`
**File**: `app/src/main/jni/jni_utils.cpp`

## Expected Behavior

The code should handle this case properly to avoid unexpected errors or degraded quality.

Nam0101:
### Problem
`init_methods_cache` uses a plain static boolean (`methods_initialized`) as a guard without synchronization. If two threads enter concurrently, one thread can observe `methods_initialized` before all global class/method IDs are fully assigned, causing later JNI calls to use null or stale `jclass`/`jmethodID` values. In native code this can lead to hard crashes (SIGSEGV/abort) during `CallStatic*Method`.

**Severity**: `high`
**File**: `app/src/main/jni/jni_utils.cpp`

### Solution
Make initialization one-time and thread-safe with `std::once_flag`/`std::call_once`, and only mark success after all lookups succeed. Example: `static std::once_flag init_flag; static bool init_ok = false; std::call_once(init_flag, [&]{ init_ok = do_init(env); });` where `do_init` validates every `FindClass`/`Get*MethodID` result before returning true.

### Changes
- `app/src/main/jni/jni_utils.cpp` (modified)

## AI or LLM usage

Closes #1163

### DIFF
--- a/app/src/main/jni/jni_utils.cpp
+++ b/app/src/main/jni/jni_utils.cpp
@@ -3,6 +3,7 @@
 
 #include <jni.h>
 #include <stdlib.h>
+#include <mutex>
 
 bool acquire_jni_env(JavaVM *vm, JNIEnv **env)
 {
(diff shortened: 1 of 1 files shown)


damontecres:
Hi, thanks for contribution. It looks like you didn't fill out the PR template fully, please update the PR body to match the template.

damontecres:

PR #1166, opened 2026-03-31
Quality: JNI class/method lookups are unchecked, allowing null IDs to be cached and crash later

### LINKED ISSUE #1165: fix: jni class/method lookups are unchecked, allowing null ids to be cached and crash later
## Description

`FindClass`, `GetMethodID`, and `GetStaticMethodID` results are assigned directly with no null checks and no handling of pending JNI exceptions. If any signature/class name is wrong (e.g., release minification/proguard mismatch, API changes), null handles are cached and used later, which typically crashes the process in subsequent JNI method invocations.

**Severity**: `high`
**File**: `app/src/main/jni/jni_utils.cpp`

## Expected Behavior

The code should handle this case properly to avoid unexpected errors or degraded quality.

Nam0101:
### Problem
`FindClass`, `GetMethodID`, and `GetStaticMethodID` results are assigned directly with no null checks and no handling of pending JNI exceptions. If any signature/class name is wrong (e.g., release minification/proguard mismatch, API changes), null handles are cached and used later, which typically crashes the process in subsequent JNI method invocations.

**Severity**: `high`
**File**: `app/src/main/jni/jni_utils.cpp`

### Solution
Validate each lookup immediately and fail initialization if any lookup returns null. Also clear or propagate JNI exceptions. For example, wrap each call: `if ((id = env->GetStaticMethodID(...)) == nullptr) { env->ExceptionDescribe(); env->ExceptionClear(); return false; }` and change `init_methods_cache` to return `bool` so callers can abort safely when initialization fails.

### Changes
- `app/src/main/jni/jni_utils.cpp` (modified)

## AI or LLM usage

Closes #1165

### DIFF
--- a/app/src/main/jni/jni_utils.cpp
+++ b/app/src/main/jni/jni_utils.cpp
@@ -16,37 +16,79 @@ bool acquire_jni_env(JavaVM *vm, JNIEnv **env)
 // Apparently it's considered slow to FindClass and GetMethodID every time we need them,
 // so let's have a nice cache here.
 
+static bool clear_pending_exception(JNIEnv *env)
+{
+    if (!env->ExceptionCheck())
+        return false;
+
+    env->ExceptionDescribe();
+    env->ExceptionClear();
+    return true;
+}
+
 void init_methods_cache(JNIEnv *env)
 {
     static bool methods_initialized = false;
     if (methods_initialized)
         return;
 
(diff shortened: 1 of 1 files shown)


damontecres:
Hi, thanks for contribution. It looks like you didn't fill out the PR template fully, please update the PR body to match the template.

damontecres:

PR #1187, opened 2026-04-02
Content Warning pill overlay on playback start

jkrishna289:
### What does this do?

Adds a small animated pill in the top-left corner of the screen that appears a few seconds after playback starts. It shows the content rating (e.g. `18+`, `13+`) and any content descriptors (e.g. `Disturbing Content · Graphic Violence · Language`).

The pill fades in, expands, holds for ~7 seconds, then quietly disappears. It shows **once per playback session** and never re-triggers on pause/resume.

---

### How does it work?

The pill reads `CW:` prefixed tags from the Jellyfin item (e.g. `CW:Violence`, `CW:Language`). These tags are already present on the item object fetched by `PlaybackViewModel.init()` — so there are **zero extra API calls**.

The rating badge is color-coded:
- 🔵 Blue — TV-Y, G, U (kids)
- 🟢 Green — TV-PG, PG
- 🟠 Amber — TV-14, PG-13
- 🟠 Orange — TV-MA, R
- 🔴 Red — NC-17, 18

Rather than showing raw strings like `TV-MA` or `NC-17`, ratings are normalized to a simple age label that anyone can understand regardless of region:

| Shows | Instead of |
|---|---|
| All | TV-Y, G, U |
| 7+ | TV-Y7, PG |
| 13+ | TV-PG, TV-14, PG-13, 12A |
| 16+ | 15, R |
| 18+ | TV-MA, NC-17, 18 |

If the rating is unknown it falls back to the raw string.

### Files changed

| File | Change |
|---|---|
| `ContentWarningPill.kt` | New — animated Jetpack Compose composable |
| `PlaybackViewModel.kt` | Extracts `CW:` tags from item, exposes as `StateFlow` |
| `PlaybackPage.kt` | Adds `ContentWarningPill` overlay inside the playback `Box` |

---

### Data source

Works with any Jellyfin setup that uses `CW:` prefixed tags. Designed to pair with the [Jellyfin Content Warnings Plugin](https://github.com/jkrishna289/Jellyfin-Content-Warnings-Plugin) which auto-tags libraries using Groq AI, but has **no hard dependency** on it. No ToS violations. 
Additionally [Jellyfin Content Warnings Plugin](https://github.com/jkrishna289/Jellyfin-Content-Warnings-Plugin) do infact have bugs but the main job works fine.

**Happy to submit a PR if this is something you'd consider merging!**

---

### Related issues
https://github.com/damontecres/Wholphin/issues/528#issuecomment-4178127886

## Screenshots

https://github.com/user-attachments/assets/981d666e-c5e5-4987-8880-dac22f1e9a15

## AI or LLM usage

> Developed with assistance from **Claude Sonnet 4.6** (Anthropic) for Compose animation design and architecture.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/playback/ContentWarningPill.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/playback/ContentWarningPill.kt
@@ -1,18 +1,23 @@
 package com.github.damontecres.wholphin.ui.playback
 
+import androidx.compose.animation.AnimatedVisibility
 import androidx.compose.animation.core.Spring
-import androidx.compose.animation.core.animateDpAsState
-import androidx.compose.animation.core.animateFloatAsState
 import androidx.compose.animation.core.spring
 import androidx.compose.animation.core.tween
+import androidx.compose.animation.fadeIn
+import androidx.compose.animation.fadeOut
+import androidx.compose.animation.scaleIn
+import androidx.compose.animation.scaleOut
+import androidx.compose.animation.slideInHorizontally
+import androidx.compose.animation.slideOutHorizontally
 import androidx.compose.foundation.background
 import androidx.compose.foundation.layout.Box
 import androidx.compose.foundation.layout.Row
 import androidx.compose.foundation.layout.Spacer
+import androidx.compose.foundation.layout.defaultMinSize
 import androidx.compose.foundation.layout.height
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.width
-import androidx.compose.foundation.layout.wrapContentWidth
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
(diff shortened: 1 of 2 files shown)


jkrishna289:
https://github.com/user-attachments/assets/31b03615-fc26-4b97-b625-6f124036ea35

damontecres:
Thanks for the contribution. However, the movie & series pages already show the rating, if available. If you think this could be shown more prominently somehow, please start an new discussion.

Furthermore, this PR seems pretty specific or tied to your new plugin. I generally try to avoid having Wholphin features be dependent on server plugins especially brand new ones.

jkrishna289:
Thanks for the feedback!

Just to clarify on both points:

1. **The rating on movie/series pages**, you're right that it's already shown there, but the pill appears *during playback* (3 seconds after the video starts), which is a different moment entirely. The idea is more like Netflix's content advisories that play at the start of a video, as a heads-up right before you're actually watching.

2. **Plugin dependency** , the pill has zero hard dependency on the plugin. It reads `CW:` prefixed tags from the item's existing tag list, which is already fetched by `PlaybackViewModel`. Any Jellyfin user can add `CW:` tags manually. The plugin I mentioned is just one way to populate those tags automatically.

That said, I completely understand if this isn't the direction you want to take Wholphin. Would a Discussion be a better place to explore whether something like this could fit. Perhaps as an opt-in setting?

damontecres:
> Would a Discussion be a better place to explore whether something like this could fit

Yup, that's the best place to discuss an idea like this and get feedback. I've been persuaded before to change my mind after reading feedback from users.

PR #1194, opened 2026-04-04
Add stats HUD to player

baykatre:
Added a stats overlay to the player. You can toggle it from the more menu, it's a separate option from the existing debug info. It shows the basics: which player engine is being used, video/audio codec details, whether it's direct playing or transcoding, decoder info, and how much buffer you have.

No new API calls or dependencies, it just uses the playback info that's already there.

Disclosure: AI-assisted (Claude). I reviewed and tested all changes.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackControls.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackControls.kt
@@ -90,6 +90,8 @@ import kotlin.time.Duration.Companion.seconds
 sealed interface PlaybackAction {
     data object ShowDebug : PlaybackAction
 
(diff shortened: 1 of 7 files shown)


Pentaphon:
Got a screenshot?

damontecres:
What's the use case for this? It shows the same info as the debug overlay.

Pentaphon:
> What's the use case for this? It shows the same info as the debug overlay.

I think he's going for the same feature as in Jellyfin-Android or the "stats for nerds" feature in Youtube where you have a compact block of info. If it's the same stats as the current overlay, though, I wouldn't blame you for closing as won't add. Or maybe you can have this one as a "simple view" and the debug overlay as the "debug view"? This is why people need to post screenshots for UI concepts.

PR #1228, opened 2026-04-10
Add a small glow effect to the segment skip button

damontecres:
## Description
Adds a small glow to the border of the segment skip buttons, ie "Skip Intro", etc. Since this button appears as focused immediately, there's no contrast and it can be missed sometimes. So this glowing border draws the eye a bit better.

Also a small change to reduce the minimum skip forward amount to 5 seconds .

### Related issues
Skip forward: #1203

### Testing
Emulator. But this needs some "real world" testing on a TV, too, so this may need tweaking

## Screenshots

https://github.com/user-attachments/assets/5488d468-aee4-4448-a823-6326ccb003da

## AI or LLM usage
None

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/playback/SkipSegmentButton.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/playback/SkipSegmentButton.kt
@@ -6,23 +6,22 @@ import androidx.compose.animation.core.LinearEasing
 import androidx.compose.animation.core.RepeatMode
 import androidx.compose.animation.core.rememberInfiniteTransition
 import androidx.compose.animation.core.tween
-import androidx.compose.foundation.BorderStroke
 import androidx.compose.foundation.interaction.MutableInteractionSource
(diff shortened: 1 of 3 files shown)


app/src/main/java/com/github/damontecres/wholphin/ui/playback/SkipSegmentButton.kt
```diff
@@ -0,0 +1,93 @@
...
+            ClickableSurfaceDefaults.border(
+                focusedBorder =
+                    Border(
+                        border = BorderStroke(width = 4.dp, color = color),
+                        inset = 0.dp,
+                        shape = CircleShape,
+                    ),
+            ),
+    )
+}
```
rozPierog: this will cause recomposition on every frame

Image

Consider using graphicLayer to create the border to avoid that

damontecres: Good call. I switched it to use `drawBehind` instead.

PR #1256, opened 2026-04-16
Add support for external player playback

### LINKED ISSUE #85: [FEA] - Support external player

damontecres:
## Description
Adds a new playback "backend" to play media in another app such as VLC.

By default, Wholphin will use the system default app. Typically if you haven't chosen one, a dialog will show to pick which app. You can also set a specific external app to use in Wholphin independent of the system default.

Currently, this only support single item playback, ie playing next up episodes is not supported yet.

Also since there is no standardized way to send resume position, external subtitle info, etc, Wholphin makes a best effort. VLC, mpv-android, & MX-Player are specifically tested/supported.

### Related issues
Closes #85

### Testing
Emulator & nvidia shield w/ VLC, mpv-android, & MX-player

## Screenshots
image

## AI or LLM usage
None

### DIFF
--- a/app/build.gradle.kts
+++ b/app/build.gradle.kts
@@ -183,6 +183,10 @@ android {
             isIncludeAndroidResources = true
         }
     }
+
+    lint {
+        disable.add("MissingTranslation")
+    }
 }
 
 protobuf {
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -30,6 +30,10 @@
         <intent>
             <action android:name="android.speech.action.RECOGNIZE_SPEECH" />
         </intent>
+        <intent>
+            <action android:name="android.intent.action.VIEW" />
+            <data android:mimeType="video/*" />
+        </intent>
     </queries>
 
     <application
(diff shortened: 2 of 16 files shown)


tunneltiger:
It’s great to see progress on external player support in this PR.

One additional suggestion:
Would it be possible to explicitly support Vimu Player integration?

Vimu is currently the only player on Sony Android TVs that can handle Dolby Vision (Profile 8.1 / 8.6) playback smoothly without stuttering, so proper integration would be very valuable.

It would be helpful if the integration ensured:

correct movie title is passed
resume position (playback timestamp) is supported

As a reference, the Jellyfin Android TV client already supports Vimu correctly, including title and resume handling. The implementation may serve as a useful reference:
https://github.com/jellyfin/jellyfin-androidtv/blob/bfd26387ba48ff26704d77abb4c3bcb130585f5d/app/src/main/java/org/jellyfin/androidtv/ui/playback/ExternalPlayerActivity.kt#L64

Vimu API:
https://vimu.tv/player-api/

Z4UBERW4LD:
Concur with tunneltiger, VIMU remains the only AndroidTV player that actually plays absolutely everything even edge cases without problems. If I could use VIMU inside Wholfin I think it’d be the perfect solution.

z3huti:
+1 for vimu player

damontecres:
I've added the necessary info for vimu based on their specs, but it is untested.

tunneltiger:
Thanks a lot for adding VIMU support—really appreciate it. Let me know if there’s a build you’d like me to test.

rickselest:
Thank you!!!!

Issue #1261, opened 2026-04-17
"Recently Released Series" row includes unplayable, unreleased, and ungrouped items


ADT95:
### Description

The "Recently Released" series row on the Wholphin home page, and the equivalent
rows on the TV and Movies sub-sections, include items that
aren't actually playable or aren't even episodes/movies:

1. Future dated episodes that Jellyfin has metadata for (via TVDB/TMDB) but haven't yet aired.
2. Episodes whose entry exists in Jellyfin but whose file isn't on disk (typical Sonarr workflow where metadata is created ahead of download).
3. On the configured home page row specifically: Season items are often rendered as if they were episodes, labelled "E<season-number>". Selecting one navigates to the season landing page rather than an episode.

There's also a secondary issue: when multiple episodes share the same release date, the release date order is random and same-series episodes are scattered across the recently released row.

The row should only contain items that are actually playable — i.e. whose file is on disk and whose premiere date has passed — and items should be limited to episodes and movies. Within the same premiere date, same-series episodes should cluster together with the latest episode first.

I have a patch for this ready to PR - happy to submit once above is acknowledged.

### Reproduction steps

1. Connect to a Jellyfin server with a Sonarr-managed TV library.
2. Make sure the library has an upcoming episode whose metadata has been ingested but whose file hasn't downloaded yet (most currently-airing series will have this).
3. Add a "Recently Released" row for that series / tv library to the home page.
4. Observe unreleased/unavailable items at the top of the row.

### App Version

v0.6.2

### Server Version

10.11.8

### Device

Chromecast with Google TV, & Nvidia Shield TV Pro

### Logs

```bash

```

damontecres:
See also https://github.com/damontecres/Wholphin/issues/1254

ADT95:
> See also [#1254](https://github.com/damontecres/Wholphin/issues/1254)

Submitted a PR with the appropriate fix - I have tested this on several devices now, with different Jellyfin libraries, and it fixes the various issues.

Issue #1254, opened 2026-04-16
[FEA/BUGFIX] - Wholphin doesn't utilize the `LocationType` response


zcraftelite9495:
## This is a megathread I made to combine together a few issues that are all related to `LocationType` processing
I am a slight expert at the backend/frontend of Jellyfin (as I manage my own fork) so I wanted to provide some help.

When you fetch the series information from the endpoint:
```http
GET /Shows/seriesId/Episodes?UserId=userId&SeasonId=seasonId
```

It returns alot of the basic values you'd expect, and more, most importantly in this case: 
```json
"LocationType": "FileSystem"
```

The issue is that Wholphin doesn't take into account that LocationType value when determining it's queue, recently added, and most of its other features' content. If the `LocationType` is `Virtual` then Wholphin should completely ignore it in some cases

I found a lot of issues on this, but all are basically due to Wholphin not handing LocationType.
To fix them:
- [ ] Wholphin should check the `LocationType` of media before adding it to the queue, if it is `Virtual` then it is unaired or unavailable, and should be ignored and not added to the queue. 
  (Should fix #1250)
- [ ] Wholphin should have a check that ignores adding anything to the "Recently Added" row if `LocationType` is `Virtual` or `Remote`. 
  (Should fix (Unaired) #262)
  (Should fix (Missing) #798)
- [ ] Wholphin should show a pill element on the top right corner (to the right of the episode indicator (`E#`)) of the inside of the episode preview identifying if it is "Unaired" (Orange) or "Missing" (Red). These should be semi-transparent and match the episode indicator's style.
  It should first check for the `LocationType` being `Virtual` and then look at the air date, marking it as `unaired` if the air date is in the future and `missing` if it is the same day or in the past. 
  An enhancement would to be adding another pill element in the bottom right corner of the inside of the episode preview that shows the air date of shows marked as `unaired` so a user can quickly see when the next episode of their favorite show airs.

### Note
I am still new to developing for Android projects, but if I do have some free time and boredom I might make a PR fix for this issue. 

### App Version

v0.6.2

### Server Version

10.11.x

### Device

Fire TV / Android Emulator

### Logs

```bash

```

PR #1262, opened 2026-04-17
Filter and reorder Recently Released Series rows

ADT95:
The 'Recently Released' rows on the home page and on the TV/Movie recommended screens previously had three issues:

1. Items appeared in the row that were not playable. The query sorted by PREMIERE_DATE descending without filtering on whether the item was actually available, so episodes whose metadata had been ingested from TVDB/TMDB but whose file was not yet on disk (typical Sonarr workflow), or episodes premiering in the future, would appear at the top of the row before being playable.
2. When multiple items shared the same premiere date (e.g. binge released seasons), the within-date order was effectively random because PREMIERE_DATE was the only sort key. Same-series episodes were scattered across the row in arbitrary order.
3. On the home page row (HomeSettingsService), the query lacked an includeItemTypes filter, so for a TV library it returned Series, Season, and Episode items mixed together. Season items were then rendered in the UI as if they were episodes (labelled 'E<season number>') but clicking them navigated to the season landing page rather than an episode.

This change adds filters, sort tiebreakers, and an item-type filter:

Filters:
- maxPremiereDate = LocalDateTime.now() - bound by current time
- isMissing = false - the file is actually on disk
- isUnaired = false - belt-and-braces for episode entities
- includeItemTypes = [EPISODE, MOVIE] on the home page row - excludes Series/Season containers so only playable leaf items appear

Sort tiebreakers:
- SERIES_SORT_NAME ascending - groups episodes from the same series adjacently within the same premiere date
- AIRED_EPISODE_ORDER descending - within a series cluster, the latest aired episode comes first (matching the descending PREMIERE_DATE semantic)
- SORT_NAME ascending (movie row only) - alphabetical tiebreaker for movies released the same day

isMissing alone is the strongest of the filters but the date and unaired filters provide additional defence against edge cases such as ingested screeners or items where the missing flag has not yet been recomputed by the server.

Closes Issue #1261 
https://github.com/damontecres/Wholphin/issues/1261

## Testing: 

Built debug APK, sideloaded on Chromecast with Google TV and Nvdia Shield TV Pro, confirmed the three bugs are fixed with specific examples (Gold Rush Season 16 Episode 1, Margot's Got Money Troubles Season 1 episodes 1-3 all related on same date correctly ordered E3→E2→E1, MasterChef Season 16 "E16" card gone.

## LLM usage:

Claude Code Opus 4.7 was utilised in fixing this bug. I used Claude to investigate the code, draft/iterate the patch, and I then tested it locally across four iterations until fully fixed and tested.


zcraftelite9495:
@damontecres Already working on a PR with almost exactly the same changes as this plus some UI improvements and fixing the other issues mentioned in #1254. I am excited to be finally taking the initiative to learn some more Kotlin, and fix some bugs ive been dealing with forever.

damontecres:
Replaced by #1436

PR #1287, opened 2026-04-20
Sort users by most recently used

Tryum:
## Description

Sort users by last recently used on User selection screen.

### Testing
Manual testing on Fire TV Stick 4k Max

## Screenshots
firetv-20260420-223109
<br>
firetv-20260420-223142

## AI or LLM usage
Claude Opus used.
Tested on device.


app/src/main/java/com/github/damontecres/wholphin/data/model/JellyfinServer.kt
```diff
@@ -58,11 +58,12 @@ data class JellyfinUser(
     val serverId: UUID,
     val accessToken: String?,
     val pin: String? = null,
+    val lastUsedAt: Long? = null,
```
damontecres, line 61: I think `Date` or `LocalDateTime` would be better. Might need a add a converter for those.

damontecres (commented):
Thanks for the contribution!

The official app has a setting to set the sort, so I think it would be a good idea to have a setting here too.

damontecres:
Just an FYI, since I had to make some databases changes in #1367, I also added the `lastUsed` field to the server & user models as a `ZonedDateTime` with a serializer.

So if you rebase on top of that, should be easier to implement this change.

Tryum:
Thanks !
I couldn´t spend time on fixing it unfortunately, but I should be able to make the change request at the end of next week.

PR #947, opened 2026-02-21
Experimental option to transcode surround sound audio to AC3

damontecres:
## Description
Adds an experimental setting to prefer transcoding audio with more than 2 channels to AC3. Stereo audio does not transcode.

This is basically a workaround for HDMI ARC (not eARC) or optical audio devices because instead of 2 uncompressed PCM channels, 6 AC3 channels (ie 5.1) is possible.

An downside of enabling this is that embedded subtitles be burned in causing a video transcode as well.

### Related issues
Fixes #255
Incorporates some code from #1410

### Testing
Just on the emulator so far to observe if the settings trigger the transcode or not

## Screenshots
N/A

## AI or LLM usage
None

PR #1290, opened 2026-04-21
Updates to the playback debug overlay

damontecres:
## Description
Lots of minor changes to the debug overlay

- Adjusts the size and layout 
- Smaller font size
- Only shows selected tracks by default
- Expanding tracks are scrollable and don't cover controls
- Show (estimated) network bitrate
- Show heap memory usage
- Better controls/chapters/queue animations
- Minor optimizations

Dev note: There's lots of code reorganization too

### Related issues
Related to #1216
Related to https://github.com/damontecres/Wholphin/issues/528

### Testing
Emulator & nvidia shield

### Selected tracks
overlay1 Large

### Expanded tracks, scrolled down
overlay2 Large

### Transcode info
overlay3 Large

## AI or LLM usage
None


DrHurt:
May I suggest wording be “DSP” when audio is offloaded from AP/CPU. It currently says “null”.

passthrough sounds like hdmi passthrough which isn’t necessarily or always the case.

Might also be worth adding (SW) & (HW) when using “c2.android…” vs “omx.mtk…” & when offloading to DSP

Edit: and if you implement tunnelled playback, that could also go there for tech savvy folks!

DrHurt:
@damontecres I see you’ve updated the wording to “non-exoplayer” which is technically correct but might be a bit confusing.

Perhaps simply calling it “hardware” or “offload” or something similar would make it clearer to end users that Wholphin is using a dedicated hardware decoder as opposed to software/CPU.

Obviously this is your app and your choice, I’m just thinking out loud as an end user. 

As always thanks for all the time and effort you put into this.

damontecres:
> I see you’ve updated the wording to “non-exoplayer” which is technically correct but might be a bit confusing

I plan to do a follow up to add more details about audio handling including offloading and device info. Probably porting some of changes in #947 where there's some debugging code to describe audio devices and HDMI ARC status, etc.

DrHurt:
Another thing I noticed is that the overlay displays the UI resolution (1080p) instead of the video pipeline display resolution (4K).

@damontecres see photo for more information 


IMG_4305

damontecres:
> Another thing I noticed is that the overlay displays the UI resolution (1080p) instead of the video pipeline display resolution (4K).

There are some quirks with getting the actual physical pixel size that `Utils.getCurrentDisplayModeSize` has workarounds for, but that's not the intention for what the debug overlay is showing. The debug overlay is showing the display mode that Wholphin chose (or the default if no choice) for playback based on the refresh rate & resolution switching settings. Maybe it needs a better label?

DrHurt:
> > Another thing I noticed is that the overlay displays the UI resolution (1080p) instead of the video pipeline display resolution (4K).
> 
> There are some quirks with getting the actual physical pixel size that `Utils.getCurrentDisplayModeSize` has workarounds for, but that's not the intention for what the debug overlay is showing. The debug overlay is showing the display mode that Wholphin chose (or the default if no choice) for playback based on the refresh rate & resolution switching settings. Maybe it needs a better label?

Better wording might help.
Technically what’s currently displayed there is UI resolution, not rendering resolution.

PR #1330, opened 2026-04-29
Accept digits in the PIN field in addition to l/r/u/d

CodeCat5:
## Description
This enhances the PIN entry system by allowing users to input digits using the numbers 0-9. This is useful for those using a keyboard or universal remote with a number pad.

### Testing
Tested on Nvidia Shield Pro and Onn 4k Pro

## Screenshots
Screenshot_20260429_083041


damontecres (approved):
LGTM, I pushed two minor UI tweaks

PR #1358, opened 2026-05-07
Show public users when switching users

### LINKED ISSUE #600: [FEA] - Autodetect existing users on selcted server and autofill username on login
### Description

When opening Wholphin for the first time and selecting an autodetected server, there should already be an existing user detected on that server and the username should be autofilled in the login prompt. Just a little QOL feature that would bring Wholphin an extra level of polish. 

This is something I see in the official WebUI and Findroid. 

Image

damontecres:
## Description
Show the public users (ie "Hide this user from login screens" is unchecked) if the server has any. Previously, all users had to be explicitly added by username or quick connect.

This PR also refactors the server & user pages to use modern state flows and deletes some unused code.

### Related issues
Closes #600

### Testing
Emulator

## Screenshots
N/A

## AI or LLM usage
None

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/MainContent.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/MainContent.kt
@@ -85,7 +85,7 @@ fun MainContent(
 
                         is SetupDestination.UserList -> {
                             SwitchUserContent(
-                                currentServer = key.server,
+                                server = key.server,
                                 Modifier.fillMaxSize(),
                             )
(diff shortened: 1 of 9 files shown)


Pentaphon:
Thank you! Does the username field in the server login prompt get autofilled with the public username when the public user is selected? Definitely nice to have instead of typing it out each time.

damontecres:
> Does the username field in the server login prompt get autofilled with the public username

Yes!

PR #1364, opened 2026-05-09
Fix:  suggestions get stuck on loading for hours/days

### LINKED ISSUE #1065: Suggestions stuck on 'Loading...'
### Description

I've noticed that at least since the last update the Suggestions category on the Recommended tab of a library just gets stuck on 'Loading...'

Image

No setting (app-side or server-side) that I see affects this. Unless I've missed one. I don't really use the Recommended tab but I thought posting this may be of some use to those who do. Although in a related 'issue' the app doesn't always remember the last tab it was on in a library, often defaulting to 'Recommended' instead of 'Library'. Some libraries it does, some it doesn't. Toggling the 'Remember Selected Tabs' setting doesn't seem to do anything.

### Reproduction steps

1. Open Wholphin
2. Navigate to a Library
3. Selected the Recommended  Tab
4. Navigate down to 'Suggested'.

### App Version

v0.5.3

### Server Version

v10.11.6

### Device

Nvidia Shield Pro 2019

### Logs

```bash

```

zoro11031:
## Description

 This should fix the issue where suggestions would sometimes get stuck loading. My best guess is that they were getting stuck because the app thought it was still loading, even though the background job was just waiting.

This should fix that by making each library kick off its own suggestions job when needed. 

  ### Related issues

  Fixes #1065

  ### Testing

Emulator and nvidia shield

  ## Screenshots

N/A no ui changes

  ## AI or LLM usage

  Used codex to help diagnosis the bug. Also added unit tests and more debug logging for the suggestions system using ai. Wrote the fix myself.


damontecres (approved):
This seems to work more reliably, especially on-demand jobs!

I am wondering if the periodic scheduling needs to be adjusted though, by either:
1) Using a unique work name per user, so a new profile signin isn't tied to same repeat interval
2) Or, use `ExistingPeriodicWorkPolicy.CANCEL_AND_REENQUEUE` instead of `UPDATE`, to replace the old user.

Since those changes aren't involved in this PR, I'm going to merge this, but it's worth thinking about.

PR #1366, opened 2026-05-10
Rearrange playback controller buttons

damontecres:
## Description
- Moves the audio track selection out of the menu as a button
- Combines the settings & more buttons into one dialog

All functionality works the same, this is just UI changes

### Related issues
Related to #528

### Testing
Emulator, it's just UI changes

### Before
controller_before Large

### After
controller_update Large

## AI or LLM usage
None

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackDialog.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackDialog.kt
@@ -43,7 +43,7 @@ import com.github.damontecres.wholphin.ui.tryRequestFocus
 import kotlin.time.Duration
 
 enum class PlaybackDialogType {
-    MORE,
+    DEBUG,
     CAPTIONS,
     SETTINGS,
     AUDIO,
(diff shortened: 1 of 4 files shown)


damontecres:
I'm wondering if the buttons should swap sides

Hukuma1:
Captions and audio being right next to each other makes sense I think. Unless you just put all 3 icons to the right.

Pentaphon:
I like it. Makes the audio setting more obvious to users upon first glance and comes second because people are more likely to change sub's than change audio tracks. Also, the 3 dot button wasn't as visible to users as a nice round settings gear is so I approve! More small changes to the player like this are welcome.

damontecres:
> Captions and audio being right next to each other makes sense I think. Unless you just put all 3 icons to the right.

Sorry, I meant possibly moving Subtitles/Audio to the left side and Settings to the right side.

Hukuma1:
I personally use audio/sub buttons more than settings so I’d rather have them on the right side. Can we see a mockup on the left side? I think it’s too busy there with all the other text.

Pentaphon:
> Sorry, I meant possibly moving Subtitles/Audio to the left side and Settings to the right side.

I would keep it as is. Keeps things consistent with other video players out there.

damontecres:
Made a mock up just to see. I definitely agree, subtitle/audio should be on the right. Thanks for the feedback!

controller_after_left Large

PR #1404, opened 2026-05-16
Fix several issues with right-to-left locales

### LINKED ISSUE #1397: Player UI in wrong orientation on RTL languages
### Description

Orientation of many UI elements are wrong when the display language has RTL layout.

### Reproduction steps

1. Switch to an RTL language (for example - Hebrew).
2. Play an item.
3. Seek bar should be from right to left but it is from left to right still, which make things I mentioned below pretty confusing
4. Player Forward/Back UI buttons are flipped -  forward should be a left arrow on the left and back should be a right arrow on the right
5. Floating menu of Sound, Subtitles and More buttons are displayed in the wrong side on the screen when opened.
6. When the first media chapter is selected below, you can't move to the next chapter.
7. Remote left/right buttons are left for backwards and right for forward, incorrectly.

Some things are OK but because the seek bar is flipped it makes them confusing:
1. Minimized seekbar on the bottom when fast-navigating is from right to left, correctly.
2. Title, time from start, time left, end time are correctly positioned from right to left.

### App Version

v0.6.4

### Server Version

10.11.8

### Device

NVIDIA Shield TV Pro 2017, Google TV Streamer 4K 2024, Homatics Box R 4K Plus

### Logs

```bash

```

### LINKED ISSUE #1398: Backdrop effect is flipped on RTL layout
### Description

When displaying backdrop of an item in the background, the effect is on the wrong side of the image.

### Reproduction steps

1. Switch to an RTL display language (like Hebrew)
2. Enter an item with a backdrop image
3. Observe that the image are correctly on displayed on the top-left side (for RTL layout) but the fade effects are also on the left side instead of on the right.

Image

### App Version

v0.6.4

### Server Version

10.11.8

### Device

NVIDIA Shield TV 2017, Google TV Streamer 4K 2024, Homatics Box R 4K Plus

### Logs

```bash

```

damontecres:
## Description
Fix several issues when the current locale is right-to-left

- Fixes the backdrop dynamic colors
- Fixes mirroring of playback controls, including seeking

### Related issues
Fixes #1397
Fixes #1398

### Testing
Emulator w/ forced RTL in developer options

## Screenshots
rtl_backdrop Large
rtl_controls Large

## AI or LLM usage

None


damontecres:
Just realized the rating star position is incorrect

damontecres:
I haven't managed to fix the rating image in the wrong place, but I don't want to hold up the rest of the fixes in this PR, so I'm going to merge it and follow up.

Issue #335, opened 2025-11-28
[BUG] - only stereo output, even though audio is 5.1


alexbudisan:
### Description

When playing videos that have support for 5.1 audio (regardless of type - EAC3, AAC, Dolby) the output is always stereo (as reported by the Sonos soundbar through which the audio plays). I’ve tried both exoplayer, as well as MPV (with and without HW decoding), but nothing. I also checked that the exoPlayer settings are correct (downmix to stereo disabled, device supports AC3/Dolby digital).

The same video plays correctly in Plex, from the same device on which Whophin is also installed (an Nvidia shield).

Is there something I’m doing wrong, or some setting somewhere that I missed? 

### Media info

Video: H.256, HEVC
Audio: ATSC A/52B (AC-3, E-AC-3), channels=6

### Player backend

ExoPlayer (default)

### App Version

0.3.2

### Server Version

10.11.3

### Device

Nvidia shield pro

### Logs

```bash

```

Pentaphon:
Does this happen on all your videos or just certain ones? I get 5.1 on my receiver when I play such DD+ files on Exoplayer.

damontecres:
Please double check if it affect AC3 and not just AAC (which is a known issue, #255).

Also, what is the "FFmpeg decoder module" setting when using ExoPlayer?

alexbudisan:
I checked several different videos, using both ExoPlayer, and MPV. Here’s what I found:

- MPV will output Stereo PCM (as reported by the Sonos soundbar) regardless of input or whether hardware decoding is turned on or off
- ExoPlayer is more complicated:
    - for videos with AAC 5.1 audio the output is always Stereo PCM
    - for videos with EAC3 or Dolby Atmos audio the output is the expected one (Dolby Digital Plus, and Dolby Atmos respectively), even supporting 7.1 channel configurations correctly

The FFmpeg decoder module setting is the default one, “Only use FFmpeg if no built-in decoder exists”.

Hope this information helps.

damontecres:
Thanks @alexbudisan!

For ExoPlayer, that's the known incorrect behavior I expect AAC 5.1 (#255).

As for MPV, I have to look, but my initial guess is that Wholphin is using/defaulting to `audio-channels=auto-safe` which falls back to down-mixing to stereo. I'll look into this.

Pabsilon:
Hey @damontecres,

Thanks for the great work!

[...]

Issue #255, opened 2025-11-19
[BUG] - AAC 6/8-channel audio only outputs as Stereo PCM


n8llcaster:
### Description

When playing video that has AAC sound with more then 2 channels - like 5.1/7.1 it will play as Stereo PCM 2.0

Bug on Jellyfin Android TV app repo: https://github.com/jellyfin/jellyfin-androidtv/issues/3900
PR's that works but haven't been merged:
https://github.com/jellyfin/jellyfin-androidtv/pull/3936
https://github.com/jellyfin/jellyfin-androidtv/pull/5025

Possible solution for most broad compatibility:
- Adding option to Advanced option to transcode all AAC audio to AC3. This codec supports every device correctly, although is only 5.1 - still better than only stereo

### Reproduction steps

1. Play video that has 5.1/7.1 AAC
2. Look at your receiver
3. You will see that app only sends PCM 2.0

### App Version

0.3.0

### Server Version

10.11.3

### Device

Sony Android TV, Nvidia Shield

### Logs

```bash

```

sudo-kraken:
Is there a plan to address this to get full 7.1 working ? This appears to have been an issue for a long time.

damontecres:
Interestingly, on my new AVR, this no longer occurs for me, I get 5.1 PCM. So this is hardware or more likely hardware capability detection related.

Some related thoughts are in https://github.com/damontecres/Wholphin/issues/335#issuecomment-3756586919

hdanielhfv:
It would be ideal if they could add the option to convert AAC and Opus 5.1 audio to AC3 5.1 audio in real time. Currently, only two Android media players can perform this action: Kodi and Vimu Media Player.

n8llcaster:
FYI **workaround** for anyone that wants this to work, just comment out AAC in supported formats and build app yourself. Works like charm.

sudo-kraken:
@n8llcaster does it work for 7.1.2?

Titaniumtown:
I am experiencing an issue like this but it actually crashes the entire application. This is on a TCL tv.

<details>

```
03-03 21:20:41.073 10661 10760 E ACodec  : [OMX.realtek.video.decoder] setPortMode on output to DynamicANWBuffer failed w/ err -2147483648
03-03 21:20:41.075 10661 10760 I ACodec  : codec does not support config priority (err -1010)
[...]

PR #1410, opened 2026-05-18
Add 'SPDIF/ARC Mode' toggle and functionality

darkflame91:
## Description
This change provides a toggle that forces all multichannel audio to be transcoded to AC3 5.1. 
This change does not force video transcode, subtitle burn-in, or alter any existing video stream transcoding behavior.

Similar to the `downmixStereo` and `ac3Supported` toggles, this new toggle applies only to Exoplayer and not mpv. 

This is meant for TV's/devices that support SPDIF and/or ARC, but not eARC. Therefore they are limited to 6 channels and the AC3 codec (many such devices support DTS codec as well, but not as universally as AC3). 
The main issue is that such devices frequently advertise support for more channels and/or newer codecs, 
but silently downmix such audio streams to stereo internally. 
This change ensures that the client will receive multichannel audio that is compatible with the device, while continuing to DirectPlay stereo audio without transcoding.
Of course, it requires audio transcoding to be enabled for the Jellyfin user and library.

This change primarily:
1. Adds a toggle to enable this mode.
2. Manages potentially conflicting 'Downmix to Stereo' and 'AC3 Supported' toggles.
3. For video libraries, if toggle is enabled, sets max channels to 6 and invalidates all non-AC3 codecs for multichannel audio.
4. For music libraries, if toggle is enabled, sets max channels to 6 and requests transcode to AC3 in mka container for multichannel audio.
5. Handles switching audio streams mid-play (taken from PR #947) 

### Related issues
#335 , #255
Similar to Experimental PR #947 - but includes (i) transcoding only if required (ii) handling multichannel music.

### Testing
1. UT's added to validate and prevent potential toggle conflicts.
2. Manually tested on Firestick 4k (1st gen) in video+audio libraries:

| Audio Format | Toggle OFF | Toggle ON |
| --- | --- | --- |
| Any stereo audio | DirectPlay | DirectPlay (unchanged) |
| AC3 5.1 | DirectPlay | DirectPlay (unchanged) |
| Any non-AC3 5.1/7.1 (incl. AAC, DTS, DD+, Atmos, DTS:X) | DirectPlay (but output as stereo) | DirectStream (only audio transcoded, output in 5.1) |
3. Verified that only audio was transcoded in Jellyfin (and ffmpeg) logs.
4. Verified that switching between stereo and multichannel audio switches between DirectPlay and DirectStream.

## Screenshots
N/A

## AI or LLM usage
LLM's were used for planning, initial draft and critical reviews. 
All code changes were manually reviewed and vetted.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/preferences/AppPreference.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/preferences/AppPreference.kt
@@ -391,7 +391,10 @@ sealed interface AppPreference<Pref, T> {
                 defaultValue = true,
                 getter = { it.playbackPreferences.overrides.ac3Supported },
                 setter = { prefs, value ->
-                    prefs.updatePlaybackOverrides { ac3Supported = value }
+                    prefs.updatePlaybackOverrides {
+                        ac3Supported = value
+                        if (!value) spdifArcSurroundAudio = false
+                    }
                 },
                 summaryOn = R.string.enabled,
                 summaryOff = R.string.disabled,
@@ -402,7 +405,10 @@ sealed interface AppPreference<Pref, T> {
                 defaultValue = false,
                 getter = { it.playbackPreferences.overrides.downmixStereo },
                 setter = { prefs, value ->
-                    prefs.updatePlaybackOverrides { downmixStereo = value }
+                    prefs.updatePlaybackOverrides {
+                        downmixStereo = value
+                        if (value) spdifArcSurroundAudio = false
+                    }
                 },
                 summaryOn = R.string.enabled,
                 summaryOff = R.string.disabled,
@@ -453,6 +459,28 @@ sealed interface AppPreference<Pref, T> {
                 },
                 summaryOn = R.string.enabled,
                 summaryOff = R.string.disabled,
+            )
+
+        val SpdifArcSurroundAudio =
+            AppSwitchPreference<AppPreferences>(
+                title = R.string.spdif_arc_surround_audio,
+                defaultValue = false,
+                getter = { it.playbackPreferences.overrides.spdifArcSurroundAudio },
+                setter = { prefs, value ->
+                    if (value) {
+                        prefs.updatePlaybackOverrides {
+                            spdifArcSurroundAudio = true
+                            ac3Supported = true
+                            downmixStereo = false
+                        }
+                    } else {
+                        prefs.updatePlaybackOverrides {
+                            spdifArcSurroundAudio = false
+                        }
+                    }
+                },
+                summaryOn = R.string.spdif_arc_surround_audio_summary_on,
+                summaryOff = R.string.spdif_arc_surround_audio_summary_off,
             )
 
         val CinemaMode =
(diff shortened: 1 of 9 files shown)


damontecres (commented):
These changes look pretty much the same as in #947. The device profile configuration and music transcoding are probably the only significant differences.

Can you elaborate more on the changes you've made?

darkflame91:
Thanks for reviewing! Yes, the changes are similar, the intents of the PR's overlap but differ, and the device profile configuration is the key difference.

My TV supports SPDIF and ARC, but not eARC. Due to SPDIF/ARC bandwidth constraints, my audio output is limited to 6 channel AC3 codec. All other multichannel audio is downmixed internally and output as stereo. I believe that such devices falsely advertise support for technically unsupported codecs and 8 channels to allow 'at least' downmixed stereo as a fallback.
Due to this, In the Wholphin release build, all videos with multichannel audio would output in stereo, unless they included an AC3 compatibility track.

From what I understood in #947 and the related issues, the intent there is to support devices that have eARC, and which either downmix or can't decode multichannel AAC and OPUS. All other multichannel audio codecs are DirectPlayed. IIRC  things like downmixing (eg. 7.1 -> 5.1) or Atmos backward compatibility with DD+ is usually smartly handled by the devices themselves.

With the #947 toggle enabled for my SPDIF usecase, I would get:
- Any Stereo -> DirectPlay (good)
- 5.1 AC3 -> DirectPlay (good)
- 5.1 AAC/Opus -> 5.1 AC3 (good)
- 7.1 AAC/Opus -> 5.1 AC3 (good) (Potentially unwanted downmix for setups with eARC and 7.1 though)
- 5.1/7.1 EAC3 -> DirectPlay downmixed to stereo (bad)
- 5.1/7.1 DTS/PCM/Other -> DirectPlay downmixed to stereo (bad)

So with this PR, the intent is to specifically support SPDIF/ARC devices by requesting audio transcode for ALL non-AC3 multichannel audio, rather than only AAC and OPUS. I'm doing this by setting the device profile to explicitly exclude all non-AC3 multichannel support if the toggle is enabled, so that AC3 is the only available compatible codec for multichannel. I've seen that other media players have an 'SPDIF mode' or similar for this reason - I've seen it in Kodi and Pot Player.

Other differences:
- I made the toggle for this feature automatically disable/enable the downmixStereo and ac3Supported toggles, with the (perhaps wrong?) assumption that any advanced user enabling this toggle would intend for those changes anyway.
- Enabled transcoding for music libraries - I have a library of multichannel music in a bunch of different formats, and this would fix playback for it. I hardcoded mka as the container for this, because it wasn't in the existing Container list and I wasn't sure if I should add it there. If container is unspecified, it defaults to mp3, and fails.
- Included the changeStreamsDirectPlay fix from #947 to make this PR feature-complete irrespective of when/if it was merged.

Can I propose a change to add an 'Audio Transcoding' sub-menu instead of these toggles, where the user can explicitly specify channel count and have granular control over the DirectPlay supported codecs?

I also ran into an issue with this PR that #947 had as well - PGS subtitles. I incorrectly said that subtitle burn-in was not forced. Turns out that since I've disabled video transcoding on my server, PGS subs were simply being dropped for me (despite 'selecting' them in the UI Subtitles menu). It looks like this is due to DirectStream remuxing to TS, which doesn't support PGS (IIRC you'd mentioned the same thing in a related thread). I'm exploring a workaround for it, but either way, I guess that would be a separate change, as it is unrelated to audio transcoding.

damontecres:
I've incorporated the differences between this PR and #947 into #947 (18d89bfc98360839c74216158afbc841d2ca7cf4), so going to close this one.

PR #1384, opened 2026-05-13
Add support for server plugin

damontecres:
## Description
PR to add support for the work-in-progress server plugin: https://github.com/damontecres/jellyfin-plugin-wholphin

### Install

[See instructions here](https://github.com/damontecres/Wholphin/releases/tag/develop-server-plugin)

### Related issues
Related to #598

### Testing
WIP

## Screenshots
N/A

## AI or LLM usage
None

PR #1418, opened 2026-05-20
Auto-configure Seerr from Wholphin server plugin

matschi95:
## Description

PR https://github.com/damontecres/jellyfin-plugin-wholphin/pull/3 defines a new endpoint for Seerr configuration in plugin. This PR consumes this new endpoint (`/wholphin/seerrsettings`) to setup Seerr on first Jellyfin login. The new plugin endpoint exposes:
- server URL
- auth method: ApiKey, Local, Jellyfin
- settings for each auth method

Jellyfin auth method supports `useCurrentUser` option, which stashes the in-memory Jellyfin password from password-based login and resuses it for automatic Seerr login. 

Methods without passwords (QuickConnect) are only pre-filling the Seerr dialog with URL and user name.  A manual configuration by entering the password is needed.

If Seerr is unreachable during the initial setup, credentials are persisted to retry Seerr login automatically at the next start of the app (so long until Seerr is reachable again).

### Related issues

- Discussion about plugin integration: https://github.com/damontecres/Wholphin/discussions/598
- Pull request for Wholphin plugin integration: https://github.com/damontecres/Wholphin/pull/1384
- **Changes needed** from jellyfin-plugin: https://github.com/damontecres/jellyfin-plugin-wholphin/pull/3

### Testing

Tested with following settings in plugin and Android TV emulator:

```yml
SeerrConfig:
  serverUrl: https://seerr.example.com
  login:
    type: Jellyfin     # ApiKey | Jellyfin | Local
    jellyfin:
      useCurrentUser: true
```

- Tested login via password and QuickConnect (only autofill)
- Tested restarts of App and restarts of Emulator
- Tested user switching

## Screenshots

no changes, only that Seerr integration gets automatically configured

## AI or LLM usage

This PR was developed in pair with Claude (Anthropic). I understand the code and can explain every change made in the PR. Manual end-to-end testing was performed by me on an Android TV emulator.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/data/model/SeerrServerPluginSettings.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/data/model/SeerrServerPluginSettings.kt
@@ -4,28 +4,28 @@ import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 
 @Serializable
-enum class SeerrPluginLoginType {
+sealed interface SeerrPluginLogin {
     @SerialName("None")
-    NONE,
+    @Serializable
+    data object None : SeerrPluginLogin
 
     @SerialName("ApiKey")
-    API_KEY,
+    @Serializable
+    data class ApiKey(
+        val apiKey: String,
+    ) : SeerrPluginLogin
 
     @SerialName("Local")
-    LOCAL,
+    @Serializable
+    data class Local(
+        val local: SeerrPluginLocalLogin,
+    ) : SeerrPluginLogin
 }
 
 @Serializable
 data class SeerrPluginLocalLogin(
-    val username: String? = null,
-    val password: String? = null,
-)
-
-@Serializable
-data class SeerrPluginLogin(
-    val type: SeerrPluginLoginType = SeerrPluginLoginType.NONE,
-    val apiKey: String? = null,
-    val local: SeerrPluginLocalLogin? = null,
+    val username: String,
+    val password: String,
 )
 
 @Serializable
--- a/app/src/main/java/com/github/damontecres/wholphin/data/model/SeerrServerPluginSettings.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/data/model/SeerrServerPluginSettings.kt
@@ -11,19 +11,9 @@ enum class SeerrPluginLoginType {
     @SerialName("ApiKey")
     API_KEY,
 
-    @SerialName("Jellyfin")
-    JELLYFIN,
-
     @SerialName("Local")
     LOCAL,
 }
-
-@Serializable
-data class SeerrPluginJellyfinLogin(
-    val useCurrentUser: Boolean = false,
-    val username: String? = null,
-    val password: String? = null,
-)
 
 @Serializable
 data class SeerrPluginLocalLogin(
@@ -35,7 +25,6 @@ data class SeerrPluginLocalLogin(
 data class SeerrPluginLogin(
     val type: SeerrPluginLoginType = SeerrPluginLoginType.NONE,
     val apiKey: String? = null,
-    val jellyfin: SeerrPluginJellyfinLogin? = null,
     val local: SeerrPluginLocalLogin? = null,
 )
 
--- a/app/src/main/java/com/github/damontecres/wholphin/data/model/SeerrServerPluginSettings.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/data/model/SeerrServerPluginSettings.kt
@@ -1,0 +1,47 @@
+package com.github.damontecres.wholphin.data.model
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+enum class SeerrPluginLoginType {
+    @SerialName("None")
+    NONE,
+
+    @SerialName("ApiKey")
+    API_KEY,
+
+    @SerialName("Jellyfin")
+    JELLYFIN,
+
+    @SerialName("Local")
+    LOCAL,
+}
+
+@Serializable
+data class SeerrPluginJellyfinLogin(
+    val useCurrentUser: Boolean = false,
+    val username: String? = null,
+    val password: String? = null,
+)
+
+@Serializable
+data class SeerrPluginLocalLogin(
+    val username: String? = null,
+    val password: String? = null,
+)
+
+@Serializable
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/github/damontecres/wholphin/data/model/SeerrServerPluginSettings.kt
```diff
@@ -0,0 +1,47 @@
+package com.github.damontecres.wholphin.data.model
+
+import kotlinx.serialization.SerialName
+import kotlinx.serialization.Serializable
+
+@Serializable
+enum class SeerrPluginLoginType {
```
damontecres: Use a sealed interface instead. This makes the code cleaner and helps prevent mistakes.

For an example, the home row config code does this.

damontecres:
I'd like to simplify this PR & https://github.com/damontecres/jellyfin-plugin-wholphin/pull/3 a bit.

Would you be okay with updating this PR so that it doesn't involve the current jellyfin user's credentials? Instead it would just read the Seerr server URL & auth type from the plugin. And if the auth type is Local or API Key, optionally, the plugin can provide those credentials that this PR reads as well.

We can discuss a more seamless jellyfin credential login separately in a follow up PR.

matschi95:
I have simplified the PR as requested by removing all Jellyfin user handling for Seerr. Therefore, automatic Seerr login is now only possible via an API key or a local Seerr account.

EDIT: I have also removed the Jellyfin Login in the plugin PR

app/src/main/java/com/github/damontecres/wholphin/ui/setup/seerr/SwitchSeerrViewModel.kt
```diff
@@ -34,6 +35,20 @@ class SwitchSeerrViewModel
 
         val serverConnectionStatus = MutableStateFlow<LoadingState>(LoadingState.Pending)
 
+        val prefilledServerUrl = MutableStateFlow<String?>(null)
+
+        fun refreshPrefilledServerUrl() {
+            if (!BuildConfig.DISCOVER_ENABLED) {
+                prefilledServerUrl.update { null }
+                return
+            }
```
damontecres: This code isn't necessary.

app/src/main/java/com/github/damontecres/wholphin/services/SeerrServerRepository.kt
```diff
@@ -50,6 +51,26 @@ class SeerrServerRepository
         val currentUser: Flow<SeerrUser?> =
             connection.map { (it as? SeerrConnectionStatus.Success)?.current?.user }
         val currentUserId: Flow<Int?> = current.map { it?.config?.id }
+        private val pluginPrefilledServerUrl = MutableStateFlow<String?>(null)
```
damontecres: I don't think this class needs to track this, it can fetch on demand

app/src/main/java/com/github/damontecres/wholphin/ui/setup/seerr/SwitchSeerrViewModel.kt
```diff
@@ -34,6 +35,20 @@ class SwitchSeerrViewModel
 
         val serverConnectionStatus = MutableStateFlow<LoadingState>(LoadingState.Pending)
 
+        val prefilledServerUrl = MutableStateFlow<String?>(null)
```
damontecres: This can be a `DataLoadingState<String>` so the dialog can wait for it to be fetched before showing the prefilled URL.

Otherwise, if there is a delay, the user could be typing and the URL will suddenly change.

app/src/main/java/com/github/damontecres/wholphin/ui/setup/seerr/AddSeerrServerDialog.kt
```diff
@@ -22,6 +22,7 @@ import com.github.damontecres.wholphin.util.LoadingState
 @Composable
 fun AddSeerServerDialog(
     currentUsername: String?,
+    currentUrl: String?,
```
damontecres: Use non-nullable or `DataLoadingState` as suggested elsewhere

app/src/main/java/com/github/damontecres/wholphin/services/UserSwitchListener.kt
```diff
@@ -77,45 +80,86 @@ class UserSwitchListener
...
-                                        )
-                                        seerrServerRepository.error(server, seerrUser, ex)
-                                    }
-                                }
-                            }
+                    launchIO { restoreOrAutoSetupSeerr(user) }
+                }
+            }
+
+        private suspend fun restoreOrAutoSetupSeerr(user: JellyfinUser) {
```
damontecres: I think this function and `tryAutoSetupFromPlugin` should be moved into `SeerrServerRepository`

app/src/main/java/com/github/damontecres/wholphin/services/UserSwitchListener.kt
```diff
@@ -77,45 +80,86 @@ class UserSwitchListener
...
+                    return
+                } ?: return
+            val url = settings.serverUrl
+            if (!url.isNotNullOrBlank()) return
+            val login = settings.login
+            if (login == null) {
+                Timber.i("Pre-filling Seerr URL from plugin")
+                seerrServerRepository.prefillFromPlugin(url)
+                return
+            }
```
damontecres: No needed if removing after removing the state from `SeerrServerRepository`.

app/src/main/java/com/github/damontecres/wholphin/services/UserSwitchListener.kt
```diff
@@ -77,45 +80,86 @@ class UserSwitchListener
...
+            try {
+                when (login.type) {
+                    SeerrPluginLoginType.API_KEY -> {
+                        val key = login.apiKey
+                        if (!key.isNotNullOrBlank()) return
```
damontecres: When you update to use a [sealed interface](https://github.com/damontecres/Wholphin/pull/1418/changes#r3449081082), you won't need these not null checks for the api key or username/password.

damontecres (changes requested):

matschi95:
I have implemented all the requested changes in the new commit.

damontecres (approved):
I think this looks good! I need to review the plugin side again and then I'll merge both PRs.

PR #1422, opened 2026-05-21
Fix bottom padding in playback controls

### LINKED ISSUE #1396: Controls have large bottom padding if there are no chapters or queue
### Description

If you play a single item that doesn't have chapters and won't have queue, there is a large blank space below the player controls.

It even looks like more space than the "Chapter" label uses.

### Reproduction steps

_No response_

### App Version

v0.6.4

### Server Version

10.11.8

### Device

_No response_

### Logs

```bash

```

JustinZeus:
Fixes #1396.

The else branch of `when (nextState)` in PlaybackController added a 32dp Spacer that left empty space below the controls when there were no chapters or queue.

Possible follow-up: Pentaphon's idea of a "similar items" row in that empty slot. Happy to put up a separate PR if useful.

## Testing

Verified on an Android TV (1080p) AVD running API 34 by playing a movie with no chapters and no next-up item. With the fix, the bottom of the control overlay sits the expected 8dp from the screen edge (the existing parent padding). Also rebuilt without the fix and compared back-to-back to confirm the Spacer was the cause.

The CHAPTERS and QUEUE branches were not modified.

## Screenshot
case1-controls

## AI or LLM usage

I used Claude to help identify the file containing the bug and to suggest the fix. I reviewed the diff before committing.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/playback/overlay/PlaybackController.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/playback/overlay/PlaybackController.kt
@@ -10,9 +10,7 @@ import androidx.compose.foundation.interaction.collectIsFocusedAsState
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
-import androidx.compose.foundation.layout.Spacer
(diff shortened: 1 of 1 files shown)


damontecres (approved):

damontecres:
Thanks for the contribution!

PR #1423, opened 2026-05-21
Add option to hide the voice search button

JustinZeus:
## Description

Adds a "Show voice search button" toggle to the search page's view options dialog, alongside the existing "Combined search results" toggle. When disabled, the voice search button is hidden from the search row, taking it out of the D-pad focus path.

The new field `show_voice_search_button` is added to `SearchPreferences` (default `true`). The serializer applies the default for new installs, and a new entry in `AppUpgradeHandler` sets it to `true` for users upgrading from earlier versions so existing behaviour is preserved.

### Related issues

Discussion: https://github.com/damontecres/Wholphin/discussions/1419

### Testing

Tested on an Android TV (1080p) AVD running API 34:

- Toggle on (default): voice search button visible at the top of the search row.
- Toggle off: button disappears; only the search text field remains.
- Toggle back on: button reappears.
- Upgrade handler verified by spoofing `version.current.name` to an older value in SharedPreferences and force-restarting the app; after `needUpgrade()` fired, the toggle was on as expected.

The upgrade-handler entry currently uses `0.6.4-8-g0` (based on current `main` HEAD); happy to bump this if you'd prefer a different target version on merge.

## Screenshots

pr1-01-home
pr1-02-toggled-off

## AI or LLM usage

I used Claude to scaffold the proto/serializer/upgrade-handler/dialog changes, mirroring the existing `combined_search_results` pattern. I reviewed each diff before committing and tested the toggle manually on the emulator, including the upgrade-handler simulation.


app/src/main/java/com/github/damontecres/wholphin/ui/main/SearchPage.kt
```diff
@@ -508,13 +524,15 @@ fun SearchPage(
...
                         .focusRequester(focusRequesters[SEARCH_ROW]),
             ) {
-                VoiceSearchButton(
-                    onSpeechResult = { spokenText ->
-                        query = spokenText
-                        triggerImmediateSearch(spokenText)
-                    },
-                    voiceInputManager = viewModel.voiceInputManager,
-                )
+                if (voiceSearchButtonVisible) {
```
damontecres: Use `AnimatedVisibility` that shrinks horizontally to start instead so the transition looks nicer.

Here's a (vertical) example you can adapt: https://github.com/damontecres/Wholphin/blob/main/app/src/main/java/com/github/damontecres/wholphin/ui/nav/NavDrawer.kt#L363-L366

app/src/main/res/values/strings.xml
```diff
@@ -449,6 +449,9 @@
     <string name="combined_search_results">Combined search results</string>
     <string name="combined_search_results_on">Show all results in a single list</string>
     <string name="combined_search_results_off">Show results by category</string>
+    <string name="show_voice_search_button">Show voice search button</string>
+    <string name="show_voice_search_button_on">Visible</string>
+    <string name="show_voice_search_button_off">Hidden</string>
```
damontecres: Might as well use a generic name because I'm sure they will be useful other places too!

```suggestion
    <string name="visible_ui">Visible</string>
    <string name="hidden_ui">Hidden</string>
```

damontecres (commented):
Thanks for the contribution! Left a couple minor comments

JustinZeus:
Thanks for the suggestions. Is this what you had in mind? @damontecres

damontecres (approved):
LGTM, I pushed a tweak to the animation.

~There's also a focus problem, but it'll require some rewriting, so I'll do it in a follow up PR~ Edit: actually it was a lot easier than I expected, so I pushed the fix

JustinZeus:
Thanks! I'm new to Kotlin and Android development. This project has amazed me in it's quality and prompted me to switch from Plex to Jellyfin. I'm a tech lead by day (web dev, CI/CD), so I felt obligated to do my part and try to help out. Thanks for the patience and acceptance.

damontecres:
I'm glad you like it! And super happy you're making it even better!

Since you're new to kotlin & android, let me elaborate on the changes I made.

First, I just think the fade and shrinking towards End looks better.

Second, moving the padding to the individual row elements instead using the `Row`'s `spacedBy` means when the voice button disappears, the `Row` doesn't abruptly have to drop the spacing between the elements. Instead the padding shrinks/expands with the button!

Finally, the focus issue happened when opening the search page if you have the voice button hidden, focus might jump to the nav drawer. This is because initially the button is always shown (the initial value was true) and then it's updated, so it focused on the button which then disappears so the focus is lost. Instead, it can use the initial value from the `userPreferences` (which will be up to date) passed in and collect any further changes to that.

Hopefully all that makes sense

JustinZeus:
Thanks for the explanations! Funny how quickly you managed to find and fix a bug like that.

PR #1431, opened 2026-05-21
Add server-plugin-defined nav pages

matschi95:
## Description

Plugin-PR https://github.com/damontecres/jellyfin-plugin-wholphin/pull/5 adds the option to define additional pages for the navigation drawer. Wholphin uses these two new endpoints in following sequence:
- when user switch happens:
  -  the client calls `/wholphin/pages` and imports these to the nav drawer layout
- when entering one of those additional pages:
  -  our client calls `/wholphin/pages/{id}` with the corresponding id
  - afterwards the client fetches the items of each row defined for this page and renders it on the screen

To avoid re-fetching every time the user navigates to a page, a small CustomPageRowsCache singleton holds the latest fetched rows per (userId, pageId). The cache is cleared on user switch.

### Known limitation: 

Plugin configuration changes don't fully propagate to the client without a logout / app restart.

The Wholphin client refreshes the navigation drawer and the home-page settings only on user switch. Custom-page row contents are fetched fresh on every page open, so row changes within an existing page show up after a brief background refresh. But anything that affects the drawer or home-screen layout is cached in-memory until the user logs out: 
- New page added in the plugin YAML → won't appear in the drawer until logout/restart
- Page removed → still visible in the drawer; clicking it returns Page not found (404)
- Page title, icon, or position changed → drawer keeps the old value
- Home-page rows changed in HomeConfig.HomePageSettings.Rows → home screen keeps the old rows

This limitation should be fixed in a seperate PR after this one was merged.

### Related issues

- Discussion about plugin integration: https://github.com/damontecres/Wholphin/discussions/598
- Pull request for Wholphin plugin integration: https://github.com/damontecres/Wholphin/pull/1384
- **Changes needed** from our Plugin: https://github.com/damontecres/jellyfin-plugin-wholphin/pull/5

### Testing

Tested via Android Emulator and following config:
```yml
Version: 1
HomeConfig:
  HomePageSettings:
    version: 1
    rows:
      - type: ContinueWatchingCombined
SeerrConfig: {}
PagesConfig:
  version: 1
  pages:
    - id: trending-movies
      title: Trending Movies
      icon: https://fonts.gstatic.com/s/i/materialicons/trending_up/v12/24px.svg
      position: AfterHome
      rows:
        - type: ByParent
          parentId: a62972cc36089d82731a68f70b8a0a38
          recursive: false
        - type: ByParent
          parentId: ff83c8bfb2a0558e7dd6036c8f81e2b8
          recursive: false
        - type: ByParent
          parentId: 963ce5c46382e9cb27e166a76853efab
          recursive: false
    - id: trending-shows
      title: Trending Shows
      icon: Play
      position: AfterHome
      rows:
        - type: ByParent
          parentId: 66953f8622de6b3b0c150d295be2dfb7
          recursive: false
        - type: ByParent
          parentId: baa6c75944109a53913c9b78dd4d0aae
          recursive: false
        - type: ByParent
          parentId: 4e62b207b7a7b6343c76a086a21c858d
          recursive: false
```

## Screenshots

My config adds two additional pages under the home entry. Trending Movies uses an URL, where as Trending Shows uses on of our supported Material Icons (Play).

Screenshot from 2026-05-22 00-11-45

## AI or LLM usage

This PR was developed in pair with Claude (Anthropic). I understand the code and can explain every change made in the PR. Manual end-to-end testing was performed by me on an Android TV emulator.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/services/ServerPluginApi.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/services/ServerPluginApi.kt
@@ -3,6 +3,7 @@ package com.github.damontecres.wholphin.services
 import com.github.damontecres.wholphin.data.model.HomePageSettings
 import com.github.damontecres.wholphin.data.model.PageConfig
 import com.github.damontecres.wholphin.data.model.PageSummary
+import com.github.damontecres.wholphin.data.model.PagesResponse
 import com.github.damontecres.wholphin.services.hilt.AuthOkHttpClient
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.json.Json
@@ -11,6 +12,7 @@ import okhttp3.OkHttpClient
 import okhttp3.Request
 import org.jellyfin.sdk.api.client.ApiClient
 import org.jellyfin.sdk.api.client.exception.ApiClientException
+import org.jellyfin.sdk.api.client.exception.InvalidStatusException
 import timber.log.Timber
 import javax.inject.Inject
 import javax.inject.Singleton
@@ -78,7 +80,7 @@ class ServerPluginApi
                     .build()
             return okHttpClient.newCall(request).execute().use { res ->
                 if (res.isSuccessful) {
-                    json.decodeFromStream<List<PageSummary>>(res.body.byteStream())
+                    json.decodeFromStream<PagesResponse>(res.body.byteStream()).pages
                 } else {
                     Timber.w("fetchPages returned HTTP %d", res.code)
                     emptyList()
@@ -102,7 +104,7 @@ class ServerPluginApi
                     Timber.w("fetchPage(%s) returned 404", id)
                     null
                 } else {
-                    throw ApiClientException(res.code.toString() + " " + res.body.string())
+                    throw InvalidStatusException(res.code, null)
                 }
             }
         }
--- a/app/src/main/java/com/github/damontecres/wholphin/services/ServerPluginApi.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/services/ServerPluginApi.kt
@@ -1,6 +1,8 @@
 package com.github.damontecres.wholphin.services
 
 import com.github.damontecres.wholphin.data.model.HomePageSettings
+import com.github.damontecres.wholphin.data.model.PageConfig
+import com.github.damontecres.wholphin.data.model.PageSummary
 import com.github.damontecres.wholphin.services.hilt.AuthOkHttpClient
 import kotlinx.serialization.ExperimentalSerializationApi
 import kotlinx.serialization.json.Json
@@ -25,11 +27,12 @@ class ServerPluginApi
 
         private val json =
             Json {
-                ignoreUnknownKeys = false
+                ignoreUnknownKeys = true
             }
 
         companion object {
             private const val HOME_CONFIG_PATH = "homesettings"
+            private const val PAGES_PATH = "pages"
         }
 
         suspend fun public(): Boolean {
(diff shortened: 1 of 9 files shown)


app/src/main/java/com/github/damontecres/wholphin/services/ServerPluginApi.kt
```diff
@@ -63,4 +66,44 @@ class ServerPluginApi
...
+            val url = createUrl(PAGES_PATH) ?: return emptyList()
+            val request =
+                Request
+                    .Builder()
+                    .url(url)
+                    .get()
+                    .build()
+            return okHttpClient.newCall(request).execute().use { res ->
+                if (res.isSuccessful) {
+                    json.decodeFromStream<List<PageSummary>>(res.body.byteStream())
```
damontecres: Wrap the list in an object. This makes it easier to add new fields in the future that older versions of the client can ignore

app/src/main/java/com/github/damontecres/wholphin/services/ServerPluginApi.kt
```diff
@@ -63,4 +66,44 @@ class ServerPluginApi
...
+                    .get()
+                    .build()
+            return okHttpClient.newCall(request).execute().use { res ->
+                if (res.isSuccessful) {
+                    json.decodeFromStream<PageConfig>(res.body.byteStream())
+                } else if (res.code == 404) {
+                    Timber.w("fetchPage(%s) returned 404", id)
+                    null
+                } else {
+                    throw ApiClientException(res.code.toString() + " " + res.body.string())
```
damontecres: Use `InvalidStatusException` instead

app/src/main/java/com/github/damontecres/wholphin/ui/nav/NavDrawer.kt
```diff
@@ -805,3 +845,44 @@ fun navItemColor(
...
+ * - one of a small Material icon name whitelist → drawn as ImageVector
+ * - anything else / null → [Icons.Default.Star] as fallback
+ */
+@Composable
+private fun CustomPageIcon(
+    iconName: String?,
+    tint: Color,
+) {
+    val trimmed = iconName?.trim().orEmpty()
+    if (trimmed.startsWith("http://", ignoreCase = true) ||
```
damontecres: Instead of doing clean up in compose code, do this work in `fetchCustomPagesByPosition` when creating the `CustomPageNavDrawerItem`.

This composable should just be something simple like:
```
when(customPageNavDrawerItem.type){
  PageNavDrawerItemType.URL-> AsyncImage(...)
  PageNavDrawerItemType.ICON->Icon(...)
  else-> Icon(Icons.Default.Star)
}
```

app/src/main/java/com/github/damontecres/wholphin/ui/nav/NavDrawer.kt
```diff
@@ -805,3 +845,44 @@ fun navItemColor(
...
+    Icon(
+        imageVector = customPageMaterialIcon(trimmed) ?: Icons.Default.Star,
+        contentDescription = null,
+        tint = tint,
+        modifier = Modifier.size(DrawerIconSize),
+    )
+}
+
+private fun customPageMaterialIcon(name: String): ImageVector? =
+    when (name.lowercase().replace("[_\\s-]".toRegex(), "")) {
```
damontecres: Move this out of compose code

See also: https://github.com/damontecres/Wholphin/pull/1431/changes#r3330387263

damontecres (changes requested):
Needs some changes.

> Plugin configuration changes don't fully propagate to the client without a logout / app restart.
> client refreshes the navigation drawer and the home-page settings only on user switch

This is true, but I do want to add refreshing the user config and other server settings more often. But that change is probably out of scope for this PR though.

damontecres:
Also, if we're going down the route of customizing the nav drawer, I'd like to make it possible to customize in-app as well. That would have be a new feature/PR.

app/src/main/java/com/github/damontecres/wholphin/services/ServerPluginApi.kt
matschi95: I have changed the response in the plugin to this:
```c#
public class PagesResponse
{
  [JsonPropertyName("pages")]
  public IEnumerable<PageSummary> Pages { get; set; } = Array.Empty<PageSummary>();
}
```

matschi95:
I have implemented your requested changes. Do you want something more to be added to this specific PR?

PR #1522, opened 2026-06-05
Add a shortcut to chosing the version of an item

damontecres:
## Description
Based on feedback from users, many felt it was difficult to 1) tell if multiple version of an item existed and 2) choose the version they want

So this PR adds a new expandable button next to Play/Resume/Restart to choose the version. It is a shortcut to the More->Choose Version dialog, but also serves as a more prominent indicator that there are multiple versions. The button is only shown if there are 2+ versions to choose from.

Also adds displaying the bitrate in the choose version dialog just in case there are identically named versions with the same resolution.

Dev note: I also added some docs for classes & functions related to choosing a version

### Related issues
Related to #652

### Testing
Emulator

## Screenshots
choose_version_button_unfocused Large

choose_version_button Large

## AI or LLM usage
None


brh150:
Nice update.

> Also adds displaying the bitrate in the choose version dialog just in case there are identically named versions with the same resolution.

How would you feel about adding displaying runtime? Could be useful in case of theatrical vs extended/special/director's cut versions, since changing such versions doesn't change the runtime in general info row under video's title.

damontecres:
> How would you feel about adding displaying runtime? Could be useful in case of theatrical vs extended/special/director's cut versions, since changing such versions doesn't change the runtime in general info row under video's title.

Great idea! Adding this in #1607

Also, I've had fixing the quick details on my list, but weirdly it's a bit complicated. I'll file an actual issue to track it though. #1608

brh150:
Thanks!

PR #1524, opened 2026-06-05
Fix combined search results losing keyboard focus when typing

### LINKED ISSUE #1523: Keyboard loses focus as results load when typing and "combine search results" is enabled
### Description

When combined search results is enabled, when you go to type into the search bar using the onscreen keyboard, it will lose focus with each character you type as results load.

### Reproduction steps

1. Go to the search page
2.  Enable "combine search results"
3.  Begin typing

### App Version

1.0

### Server Version

10.11.10

### Device

Nvidia Shield TV Pro 2019

### Logs

```bash

```

zoro11031:
## Description
Minor tweak to prevent the keyboard from losing focus when typing search queries when "combine search results" is enabled

### Related issues
Fixes #1523

### Testing
Emulator only

## Screenshots
N/A

## AI or LLM usage
None

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/main/SearchPage.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/main/SearchPage.kt
@@ -1017,7 +1017,6 @@ private fun <T : CardGridItem> SearchGrid(
     positionCallback: (columns: Int, position: Int) -> Unit,
     modifier: Modifier = Modifier,
 ) {
-    RequestOrRestoreFocus(focusRequester)
     Column(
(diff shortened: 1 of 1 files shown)


damontecres (approved):
LGTM! Thanks!

I also pushed another focus related fix. Once the user clicks search to trigger an "immediate" search, `searchClicked` is set to true, but doesn't ever reset. So if the user then changes the query, focus is moved down like you described in your issue. So the change just resets `searchClicked` to false whenever the query is changed.

damontecres:
Also, focus bugs are the worst!

zoro11031:
Thanks for the assist! Tell me about it man, its like playing whack a mole

PR #1527, opened 2026-06-05
Improved strm file support

### LINKED ISSUE #216: [BUG] - Issue inherited from official app. When playing a strm file or any remote media for the first time, audio tracks and subtitles are not probed/toggle-able
### Description

When a .strm media source or remote media source of any kind for a movie/show is played for the first time. It falls back to whichever default language is set for the video and does not probe or allow for switching of audio tracks or subtitle tracks. When playback is quit out of and played for a second time, the audio tracks and subtitle tracks are now toggle-able and able to be switched between. It seems the android tv client/wholphin is not probing for audio/tracks or subtitles. Until playback has started once. The other clients do not seem to have this problem and allow for switching between audio/subtitle tracks upon first playback (web client, desktop client etc)

### Reproduction steps

1.Play remote media file in jellyfin with multiple audio tracks
2. Attempt to change audio track or subtitles.

### App Version

0.2.8

### Server Version

10.11.2

### Device

Firestick 4k max

### Logs

```bash

```

damontecres:
## Description
Improves support for strm files

When loading the details of a media item that is backed by the strm file, the app will query the server as if playback for it is starting. This causes the server to probe the remote source.

After that, the details are updated. The play buttons are disabled during the process as well.

### TODO
- [x] More testing
- [x] Add loading indicator while waiting for the server to get remote metadata?
- [ ] Need to validate assumptions about missing `MediaStream` data
- [x] Need to check if any playback sessions need to be closed

### Related issues
Closes #216
Requires #1522

### Testing
Emulator testing for videos, movies, & episodes

## Screenshots
N/A

## AI or LLM usage
None

## Try it out?

See [instructions here](https://github.com/damontecres/Wholphin/releases/tag/develop-strm-file)


lostb1t:
fyi: strm files can point to local files and isRemote will be false for them. (but they still need to be probed)

cmeka:
If this is automatic please consider making this optional (opt-in) or add an API URL parameter or this could completely break my use case for strm files, which is reduced server load from not analyzing every media item. If this looks like regular playback on the server then I have no way to block this feature.

damontecres:
> fyi: strm files can point to local files and isRemote will be false for them. (but they still need to be probed)

I updated the code so it doesn't check `isRemote`, but I got playback errors when I tested. Turns out starting with `10.11.7`, this is [no longer supported](https://github.com/jellyfin/jellyfin/issues/16557#issuecomment-4247853382) for security reasons.

ashipaek0:
**CAVEAT**
Testing this (WITH REMUX SERVER). 

It probes the first stream only

Doesn't let me change the subtitles in the player

Also, changes to subtitles and audio track in the 3 dots  details page don't appear to work in the player in my testing 

When I switch versions of the stream it doesn't probe the new stream chosen 

Inconsisteny with the next track in a season

I have to back out and into the season, before navigating the episodea trigger the probing

datwildfire:
> **CAVEAT** Testing this (WITH REMUX SERVER).
> 
> It probes the first stream only
> 
> Doesn't let me change the subtitles in the player
> 
> Also, changes to subtitles and audio track in the 3 dots details page don't appear to work in the player in my testing
> 
> When I switch versions of the stream it doesn't probe the new stream chosen
> 
> Inconsisteny with the next track in a season
> 
> I have to back out and into the season, before navigating the episodea trigger the probing

Omg, so close, hope this can get solved and this would be complete perfection! ❤️ thanks @damontecres, you are the goat

PR #1555, opened 2026-06-10
Add Streamystats integration

matschi95:
## Description

Streamystats (https://github.com/fredrikburmester/streamystats) is a Jellyfin companion service that tracks user watch history and provides AI-based recommendations for each user.

This PR adds a Streamystats integration to the settings dialog. The server URL is stored per Jellyfin user/server and is tested before saving. This setting is already behind a small repository/source layer, so this should make it straightforward to load the config from the Wholphin Jellyfin plugin later on. The existing Seerr integration was used as a reference for the implementation of the settings dialog.

When Streamystats is configured, Wholphin can show two new home rows:

- Recommended Movies (Streamystats)
- Recommended Shows (Streamystats)

The Streamystats rows are available in the home layout editor. If no custom home layout is configured, they are automatically added to the default home layout while the Streamystats integration is active.

Removing the integration also removes saved Streamystats rows from custom home settings.

Streamystats also exposes promoted/seasonal watchlists (Christmas, Halloween,...). I left those out of this PR because they need a separate change to how Wholphin expands home layout rows.

### Related issues

None

### Testing

Tested locally with my Jellyfin server and Streamystats instance inside an AndroidTV emulator.

Checked:

- adding a valid Streamystats URL
- rejecting invalid/unreachable URLs
- loading movie recommendations
- loading show recommendations
- adding Streamystats rows in the home layout editor
- removing the integration again
- empty recommendation results

## Screenshots

Screenshot from 2026-06-10 17-59-44

Screenshot from 2026-06-10 18-02-33

Screenshot from 2026-06-10 18-02-17

## AI or LLM usage

AI assistance was used during implementation. I reviewed the changes myself and tested them locally as described above.

### DIFF
--- a/app/schemas/com.github.damontecres.wholphin.data.AppDatabase/35.json
+++ b/app/schemas/com.github.damontecres.wholphin.data.AppDatabase/35.json
@@ -1,0 +1,795 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 35,
+    "identityHash": "00edfbdaab4d78440ff7b3b2b67194a5",
+    "entities": [
+      {
+        "tableName": "servers",
+        "createSql": "CREATE TABLE IF NOT EXISTS `${TABLE_NAME}` (`id` TEXT NOT NULL, `name` TEXT, `url` TEXT NOT NULL, `version` TEXT, `lastUsed` TEXT, PRIMARY KEY(`id`))",
+        "fields": [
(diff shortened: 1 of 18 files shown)


damontecres:
Thanks for effort, but for now, I don't want to integrate another third-party API.

Maybe I can revisit this in a few months.

Issue #1372, opened 2026-05-11
Music playback issue on TCL TV


jameswathomson:
### Description

Latest stable and beta mostly do not playback, stay at 0s. Occasionally I can get it to play via fiddling with settings i.e. press instant mix plays for a while then stops. I have not tried any previous version having just switched to Wholphin

### Media info

Audio mp3 files

### Player backend

ExoPlayer (default)

### App Version

0.6.3

### Server Version

10.11.8

### Device

TCL 55C745K

### Logs

```bash

```

damontecres:
Is it all music files?

Also, if you could [gather the app logs](https://github.com/damontecres/Wholphin/wiki/Gathering-app-logs) that could help.

PR #1564, opened 2026-06-11
Add a preference option to disable hardware audio offloading

Nehereus:
## Description
By default, Wholphin plays music with hardware audio offloading enabled. This delegates audio decoding to the device's hardware DSP, making playback stability highly dependent on the manufacturer's system audio implementation. On certain hardware (such as my TCL TVs, e.g., the Chinese Model Q9L Pro), this implementation causes music playback to hang indefinitely at 0 seconds.

### Related issues
#1372

### Testing
Tested on a TCL TV (Q9L Pro):
- Some musics playback hangs at 0s with offloading enabled.
- Disabling audio offloading fixes the issues for all medias.

## Screenshots

IMG_9486

## AI or LLM usage
Developed in pair with Gemini and human-reviewed, tested.


devwebit4u:
Having the exact same problem with my TCL C81K. Hope that this will fix the issue.

app/src/main/java/com/github/damontecres/wholphin/preferences/AppPreference.kt
```diff
@@ -956,6 +956,16 @@ sealed interface AppPreference<Pref, T> {
                     }
                 },
             )
+        val EnableAudioOffload =
```
damontecres: Could you move this to the new experimental settings?

See #1624 for an example

app/src/main/java/com/github/damontecres/wholphin/services/PlayerFactory.kt
```diff
@@ -53,6 +57,7 @@ class PlayerFactory
     constructor(
         @param:ApplicationContext private val context: Context,
         @param:AuthOkHttpClient private val authOkHttpClient: OkHttpClient,
+        private val appPreferences: DataStore<AppPreferences>,
```
damontecres: Once you merge in `main`, this won't be needed

app/src/main/res/values/strings.xml
```diff
@@ -393,6 +393,8 @@
     <string name="will_remove_pin">Will remove PIN</string>
     <string name="pin_digits_hint">[0-9]</string>
     <string name="image_cache_size">Image disk cache size (MB)</string>
+    <string name="enable_audio_offload">Enable audio hardware offload</string>
```
damontecres: I think reverse this maybe, so toggling `on` will diable audio offload instead.

damontecres (changes requested):
Thanks for the contribution!

I'd just like to move this setting to experimental and you'll need to clean up some code after merging/rebasing main.

app/src/main/java/com/github/damontecres/wholphin/services/PlayerFactory.kt
Nehereus: I’ve removed the variable by modifying the function signature. LMK if this isn’t what you mean. 
```kt
fun createAudioPlayer(
            disableAudioOffload: Boolean,
            extensions: MediaExtensionStatus = MediaExtensionStatus.MES_FALLBACK,
        ):
```

app/src/main/res/values/strings.xml
```diff
@@ -400,6 +400,8 @@
     <string name="will_remove_pin">Will remove PIN</string>
     <string name="pin_digits_hint">[0-9]</string>
     <string name="image_cache_size">Image disk cache size (MB)</string>
+    <string name="disable_audio_offload">Disable audio hardware offload</string>
+    <string name="disable_audio_offload_summary">Offload audio decoding to the hardware DSP may cause playback hangs on some devices (e.g. TCL TVs). Toggling this option will disable audio hardware offload. Restart the app to take effect.</string>
```
damontecres: I want to simplify the summary a bit

```suggestion
    <string name="disable_audio_offload_summary">Disabling offloading audio decoding may fix playback freezes on some devices</string>
```

.gitignore
```diff
@@ -63,4 +63,4 @@ app/src/main/libs/
 app/src/main/jnilibs
 app/src/main/jnilibs/
 scripts/mpv/deps/
-scripts/mpv/prefix/
+scripts/mpv/prefix/
```
damontecres: Need an empty newline to pass style checks
```suggestion
scripts/mpv/prefix/

```

damontecres (commented):
Two minor changes needed

app/src/main/java/com/github/damontecres/wholphin/services/PlayerFactory.kt
damontecres: I just meant that injecting the app preferences wouldn't be needed after the merge since the full preferences are passed into the create function, so you're good!

damontecres (approved):
Looks good, will merge once CI passes

andrewpym:
When will this be available to test.  I have a TCL TV and can confirm the issue on TCL, using google dongle and there is no issue with music playback, but the exact same setup on TCL and it fails.  The logs show Jellyfin sent the play commend and nothing happens.  So id love to be a beta tester for this.

andrewpym:
I can confirm using this option on TCL C855 and with Jellyfin 12 RC5 has now fixed the issue of music play back.

krilok:
Unfortunately, disabling hardware audio offloading still doesn't make it work on my TCL QM8B, and no music is played at all.

Edit: My bad. Once the option was toggled, I had to force close Wholphin completely and restart it again, and music playback now works. 😉

PR #1571, opened 2026-06-12
Add localized duration formatting and related string resources

tal-sarid:
## Description
Instead of using English-only duration format (e.g. `HHh MMm SSs`)  in various places, I've added a format function that is based on several new localized resources keys.

That includes:
1. Media runtime in Item view
2. Time remaining in Item view
3. End time in Player view
4. Time passed and time remained in Player view

### Related issues
None

### Testing
I installed the app on my Google TV Streamer, switched to my language, browsed all item views that I could think of to see that the duration labels are all localized in my language. I didn't commit those strings, I will do it through the Weblate after it will get merged. I also played a few items and saw that the labels there are localized as well.

## Screenshots
It looks the same in English as before, currently no other language has custom resources for it yet.

## AI or LLM usage
Used Claude Sonnet to detect places that I missed that are related to this localization change. Also it helped me to get access to my format function from various places in the code.


app/src/main/java/com/github/damontecres/wholphin/ui/detail/PlaylistDetails.kt
```diff
@@ -810,7 +812,7 @@ fun PlaylistItem(
                     }
                 Column {
                     Text(
-                        text = duration.toString(),
+                        text = LocalResources.current.formatDuration(duration),
```
damontecres: Wrap with `remember`

app/src/main/java/com/github/damontecres/wholphin/ui/playback/overlay/PlaybackControls.kt
```diff
@@ -275,16 +277,17 @@ fun SeekBar(
             horizontalArrangement = Arrangement.SpaceBetween,
         ) {
             val remaining = ((player.duration - position) / 1000).seconds
+            val resources = LocalResources.current
             Text(
-                text = (position / 1000).seconds.toString(),
+                text = resources.formatDuration((position / 1000).seconds),
```
damontecres: It would be good to wrap with a remember, keyed on `position / 1000`.

Same for below for `remaining`.

damontecres: I know the existing code doesn't, but it should have!

app/src/main/java/com/github/damontecres/wholphin/data/model/BaseItem.kt
```diff
@@ -159,10 +162,17 @@ data class BaseItem(
...
-                                        ?.let { add(it.toString()) }
+                                        ?.let { add(WholphinApplication.instance.resources.formatDuration(it)) }
                                     data.timeRemaining
                                         ?.roundMinutes
-                                        ?.let { add("$it left") }
+                                        ?.let {
+                                            add(
+                                                WholphinApplication.instance.resources.getString(
+                                                    R.string.time_left,
+                                                    WholphinApplication.instance.resources.formatDuration(it),
```
damontecres: Assign `WholphinApplication.instance.resources` to a variable so these lines can be shorter

app/src/main/java/com/github/damontecres/wholphin/data/model/BaseItem.kt
```diff
@@ -159,10 +162,17 @@ data class BaseItem(
...
                                         ?.ticks
                                         ?.roundMinutes
-                                        ?.let { add(it.toString()) }
+                                        ?.let { add(WholphinApplication.instance.resources.formatDuration(it)) }
                                     data.timeRemaining
                                         ?.roundMinutes
-                                        ?.let { add("$it left") }
+                                        ?.let {
+                                            add(
+                                                WholphinApplication.instance.resources.getString(
```
damontecres: Just an aside, this code eventually needs to be reworked. One reason I was dragging my feet to fix this string is because it won't update when switching the app's language. The app has to be closed or pages reloaded from scratch.

But it would be a significant refactor to fix that and outside the scope of this PR.

damontecres (changes requested):
Thanks for the contribution! I'm always happy to see improvements to the localization code!

I left a few comments mainly about adding `remember`s so recompositions will be a bit faster.

tal-sarid:
Did the changes you've asked. Also, I found more places to localize that I missed earlier and updated the branch to fix the merge conflicts.

damontecres (approved):
I think this looks good. I'll run some tests!

PR #1580, opened 2026-06-16
Fix AFTKRT + HSL/TS + EAC3 7.1 playback issues

calebvetter:
## Description
This is a targeting fix for a very specific bug: on Fire TV 4k Max (AFTKRT), and possibly others but untested, ExoPlayer can't play an HLS/TS stream that has 8-channel (7.1) EAC3 audio. Direct play is not affected (7.1 EAC3 plays fine), and fewer channels work fine (i.e. 5.1 EAC3) even in an HLS/TS wrapper.

This bug + device is tracked in KnownDefects. If more models are tested and found to have the same issue, they can be added to the list. Top candidates are probably AFTKA and AFTKM since they are also in the models with HEVC DoVi/HDR10+ playback issues and so probably have similar underlying limitations, but I don't have those devices to test.

The fix is to first allow the app to make a playback request to the server and receive the stream info Jellyfin has decided to send. Then it evaluates the settings and stream for the following:
- Playback is using ExoPlayer
- Device is in the modelsWithEac3HlsPassthroughBug list (currently only AFTKRT)
- Stream is not direct play
- Stream is HLS with MPEG-TS container
- Audio is 8-channel EAC3

We have to use a two-request fix because we don't know what Jellyfin is going to decide for the stream until we receive it. That way normal playable streams aren't getting overwritten by the fix when they shouldn't be. So the only streams that get the two request path applied is the exact fix for this bug.

To apply post-playback-info fixes (including this one) a helper file was created PostPlaybackInfoFixes. In the future, if there are more bugs that can only be determined after the decided stream info is received, they can be added here. This keeps things logical and maintainable.

### Related issues
The bug was able to be reproduced on multiple items with this exact scenario by simply playing back the item (making sure ExoPlayer is set as the default player in settings).

### Testing
Tested on the target device: Amazon Fire TV 4k Max (2nd Gen), model AFTKRT. 

The following media was tested:
- ExoPlayer, multiple titles with HEVC, EAC3 7.1 copied into HLS/TS (all failed, only processing 1-2fps)
- Direct play of the same titles (works)
- EAC3 5.1, HLS, MPEG-TS (works)
- MPV with failing titles (works)

The following device and settings were tested and had no effect on the above media results, ruling out external connections and settings:
- HDMI to an LG TV, both connected to a soundbar via ARC, and disconnected using internal speakers
- HDMI to a Dell computer monitor
- Fire TV Settings > Display & Sound > Audio > Surround Sound > all 4 settings tested (Best Available, PCM, Dolby Digital Plus, Dolby Digital)

## AI or LLM usage
Codex was used to debug the issue, create the initial change, and do iterative testing on the device to narrow the scope. The code was then carefully reviewed and I made structural decisions so everything fits well within the project's established conventions, removing bloat, added comments and kept it targeted, clear and maintainable.


damontecres (commented):
I think this can be implemented directly in the [device profile](https://github.com/damontecres/Wholphin/blob/main/app/src/main/java/com/github/damontecres/wholphin/util/profile/DeviceProfileUtils.kt#L135) instead.

You can add a condition to only allow transcoded EAC3 if channels <8 and add a new transcode profile for 8 channels that doesn't accept EAC3.

calebvetter:
You're totally right. Didn't realize we could get so granular on the first stream request.

I rebased onto the current upstream main, reverted PlaybackViewModel since it doesn't need to change at all, and added a simple codecProfile only when the device is in the known defects list and EAC3 is enabled.

I also tested it further by creating a standalone ts file on the server to be played without HLS to see if this broader codec profile which covers _all_ TS and not just HLS is correct. The standalone TS file also failed to playback, so not only is it a simpler fix, but the broader scope is even more accurate. I renamed the KnownDefects item and comments accordingly.

damontecres (commented):
I think this looks good, I just need to find a moment to test on my physical device!

damontecres:
Tested this and it seems to work well!

Issue #1598, opened 2026-06-20
SerializationException in `onSaveInstanceState` when the back stack contains an `ItemGrid` destination


ctismer:
### Description

When Android calls `onSaveInstanceState` while the navigation back stack contains a
`Destination.ItemGrid`, the app crashes with a `kotlinx.serialization.SerializationException`.
The grid's `@Contextual request: T` (a Jellyfin SDK request such as `GetItemsRequest`) has no
serializer registered in the `Json` instance used to persist the back stack.

### Reproduction steps

1. Navigate into any grid backed by a request (library / genre / person / filtered view → an `ItemGrid` destination).
2. Background or stop the activity (Home button, or after returning from playback) so `onSaveInstanceState` runs.
3. The app crashes (caught by ACRA, crash dialog shown).

### App Version

v0.5.1 

### Server Version

10.11.6

### Device

Amazon Fire TV Stick (AFTMM), Fire OS / Android 7.1.2 (SDK 25)

### Logs

```bash
kotlinx.serialization.SerializationException: Serializer for subclass 'GetItemsRequest' is not found in the polymorphic scope of 'Any'.
Check if class with serial name 'GetItemsRequest' exists and serializer is registered in a corresponding SerializersModule.
	at kotlinx.serialization.PolymorphicSerializerKt.findPolymorphicSerializer(PolymorphicSerializer.kt:109)
	at kotlinx.serialization.json.internal.StreamingJsonEncoder.encodeSerializableValue(StreamingJsonEncoder.kt:250)
	...
	at com.github.damontecres.wholphin.ui.nav.Destination$ItemGrid.write$Self$app(Destination.kt:112)
	at com.github.damontecres.wholphin.ui.nav.Destination$ItemGrid$$serializer.serialize(Destination.kt:112)
	...
	at com.github.damontecres.wholphin.MainActivity.onSaveInstanceState(MainActivity.kt:521)
	at android.app.Activity.performSaveInstanceState(Activity.java:1430)
	at android.app.ActivityThread.handleStopActivity(ActivityThread.java:3976)
```

PR #1599, opened 2026-06-20
Guard back-stack serialization to avoid crash in onSaveInstanceState

ctismer:
## Description
ItemGrid destinations carry @Contextual request types with no registered serializer, which threw a SerializationException when Android saved the back stack. Wrap save and restore in try/catch so persistence is best-effort: on failure the back stack is simply not persisted/restored instead of crashing.

### Related issues
Workaround for #1598. This stops the crash but does not fix the root cause — the proper fix is to register the request types and RequestHandler polymorphism in the nav Json's serializersModule (see the issue).

### Testing
Built and ran on the Android TV emulator and a Fire TV Stick (AFTMM). assembleDefaultDebug and the unit tests pass; ktlint via pre-commit is clean. Verified the app no longer crashes when backgrounding from an ItemGrid.

## Screenshots
N/A — no UI changes.

## AI or LLM usage
Disclosure: this change was written with AI assistance (Claude Code). I directed the design, reviewed the diff, and tested it on an emulator and a Fire TV Stick. I understand the code and can maintain it.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
@@ -151,10 +151,22 @@ class MainActivity : AppCompatActivity() {
         lifecycle.addObserver(playbackLifecycleObserver)
 
         val backStackStr = savedInstanceState?.getString(KEY_BACK_STACK)
-        if (backStackStr != null) {
+        val restoredBackStack =
+            if (backStackStr != null) {
+                try {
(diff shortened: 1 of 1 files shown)


damontecres:
Thanks for the contribution! I used your commit in #1605 to guard the save/restore and added the necessary serialization config to fix the root problem.

So this PR will be closed in favor of #1605

ctismer:
Thanks for the quick turnaround and the proper root fix — really happy with how this turned out.

Issue #702, opened 2026-01-14
Seerr/Discover Follow ups


damontecres:
Just task tracking for Seerr/Discover work started in #558 

- [x] Support season selection
- [x] Support quality selection #685 
- [ ] Show more suggestions (click end of row to go to paginated grid?)
- [ ] Discover by genre & other seerr front page categories
- [ ] Maybe add a way to request quality upgrades for items already in the library
- [ ] Support multiple seerr servers

### Seerr v3.x tasks

- [ ] TVDB for metadata provider

sibero80:
Is there a way to see the "processing" progress of the requested media in seerr?

NathanEmb:
Love the app. I had a request for this integration, not sure if this is the right place for it

- Can the view distinguish between Movie/Series in the "Now trending" section? It could just mimic how seerr does it today, putting a little Movie/Series badge in the top left ie: 

<img width="1203" height="420" alt="Image" src="https://github.com/user-attachments/assets/e411f218-51cf-48a0-a404-8bb2408edb2c" />

I expect to have some more freetime coming in about a month or so and would love to contribute so if this is something the wholphin team supports I might tacklet it myself as a first issue if it's still available

damontecres:
> Can the view distinguish between Movie/Series in the "Now trending" section

Done in https://github.com/damontecres/Wholphin/pull/1178

yutamago:
Wish I could filter by tags, or create filtered views.

I either search for anime, or for literally anything else.
A filtered view would make it much easier to find them.

byted:
In its current form, can one search for specific items to request? I was hoping to be able to kick the dedicated Seerr app from my TV.

damontecres:
> In its current form, can one search for specific items to request?

Yes, the regular search page will search Discover/Seerr as well.

MulverineX:
It'd be nice if Wholphin could use the Jellyfin auth to automatically log users into Seerr. (if that is the auth mode setup on the Seerr instance)

It's annoying to have log everyone in twice.

Edit: Looks like there's a PR, just no issue, https://github.com/damontecres/Wholphin/pull/1418

gionaufal:
In Seerr a user can create custom sliders/lists based on keywords, genres, studios, among others. It would be really nice if Wholphin would pick up those custom lists and show them in the UI directly. I think it would fit into this follow up issue.

[...]

PR #1614, opened 2026-06-23
Add filtering & sorting for discover grid pages

damontecres:
## Description
This PR adds sort & filter options for the discover grid pages.

Sort options:
- Popularity
- Release date
- "Community rating" (TMDB user vote)
- Name

Filter options:
- Genre
- Studio/Network
- "Community rating" (TMDB user vote)
- Parental rating
- Production status (TV only)

Also, two more tabs are added to the discover page for browsing movies and TV. These are essentially the same as clicking "View more" on the respective rows on the Discover tab.

### Limitations

This PR hardcodes the region used for studio listing and filtering to "United States". I'll work on a follow up PR to enable users to specify this.

The sort & filters are not persisted. I think they should be on the two new tabs, but not elsewhere, but this will be a follow up PR.

### Related issues
Related to #702 
Implements some of #1511
Rewrites the filter logic in #1610 

### Testing
Emulator, but needs further testing

## Screenshots
discover_filters Large

## AI or LLM usage
None


Asinin3:
@damontecres This is great, I just wish we still had the filter option for Genre's there if you've already gone into a genre.

ATM, if you go Discover > Movies > All the way to the right then view more. You get the filter option for Genre's and all the other options, which is nice because you can change genre or select multiple.

But if you go Discover > Genres in Movies > Filter. You will see that the filter options wont let you change genre or add another. It makes you have to go in and out out of the pages a bit.

PR #1619, opened 2026-06-23
Respect system 24-hour setting in clock time format

ctismer:
## Description

The clock ignored the device's system 24-hour preference: getTimeFormatter derived 12/24h purely from the locale. This reads DateFormat.is24HourFormat and builds the time pattern from it, so the clock follows the system setting.

### Testing

Tested on the emulator toggling the system 24-hour setting on/off. assembleDefaultDebug, testDefaultDebugUnitTest and pre-commit (ktlint) pass.

## AI or LLM usage

The analysis and the fix were done by me with AI assistance (Claude). I wrote the description myself; the English wording was polished with AI. I understand the code and can maintain it.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/Formatting.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/Formatting.kt
@@ -1,5 +1,7 @@
 package com.github.damontecres.wholphin.ui
 
+import android.content.Context
+import android.text.format.DateFormat
 import androidx.annotation.StringRes
 import androidx.compose.foundation.text.appendInlineContent
 import androidx.compose.ui.text.AnnotatedString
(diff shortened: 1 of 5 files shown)


app/src/main/java/com/github/damontecres/wholphin/ui/Formatting.kt
```diff
@@ -17,14 +18,9 @@ import java.time.format.DateTimeParseException
...
 
-private var timeFormatter: DateTimeFormatter =
-    DateTimeFormatter.ofLocalizedTime(FormatStyle.SHORT).withLocale(Locale.getDefault())
-
 fun getTimeFormatter(): DateTimeFormatter {
-    if (timeFormatter.locale != Locale.getDefault()) {
-        timeFormatter = timeFormatter.withLocale(Locale.getDefault())
-    }
-    return timeFormatter
+    val pattern = if (DateFormat.is24HourFormat(WholphinApplication.instance)) "HH:mm" else "h:mm a"
```
damontecres: I think this should be used instead: https://developer.android.com/reference/kotlin/android/text/format/DateFormat#gettimeformat

And I think the `Context` needs be passed in (from `LocalContext.current` in most cases) because I believe the application context might not always have the current locale if the user switches languages.

damontecres (changes requested):

app/src/main/java/com/github/damontecres/wholphin/ui/Formatting.kt
ctismer: Good call — thanks! This started as a quick fix for my own setup, but you're right it needed the general solution. Now uses DateFormat.getTimeFormat(context) and passes the Context from LocalContext.current, since the application context can carry a stale locale after an in-app language switch. Also folded TimeRemaining into EndsAt to drop the duplicated rendering. Build, tests and ktlint pass.

damontecres (approved):
Looks good, thanks!

PR #1624, opened 2026-06-24
Add experimental settings & video tunneling

damontecres:
## Description
This PR introduces the concept of "experimental" settings or features. These may cause instabilities or may need further testing. They may be changed or removed any time. Also, the developers won't provide lots of support for them.

The goal is expose some specialized settings that might only apply to a small subset of users or test out tweaks to playback.

The first experimental feature is a toggle to enable [video tunneling](https://developer.android.com/media/media3/exoplayer/track-selection#tunneling).

### Related issues
Related to #1283

### Testing
Emulator

## Screenshots
N/A

## AI or LLM usage
None


DrHurt:
Is there a way to tell if tunnelling is actually on during playback? Would be nice to have that on the debug overlay.

PR #1637, opened 2026-06-27
Add option to never automatically hide controls

damontecres:
## Description
Adds an option to disable automatically hiding playback controls. It can be enabled by by moving higher than 15 seconds.

Also, a slight UI tweaks to use a fixed width for the slider preference labels. This prevents the slider bar from changing size if the label becomes shorter or longer (eg "9 seconds"=>"10 seconds").

### Related issues
Closes #1629

### Testing
Emulator

## Screenshots
N/A


mschumacher69:
Suggestion, maybe better to add a "Never" toggle under the slider which disables the slider when enabled. I think this would be better than having to drag the slider all the way past 15 seconds to disable auto hide, considering the slider moves in 0.1 second increments. This also makes it clear that there is a "Never" option; hiding the "Never" option behind 15 seconds might make it unknown to (most) users who don't know that it exists.

damontecres:
> maybe better to add a "Never" toggle under the slider which disables the slider when enabled

I've been trying to avoid adding more and more toggles as much as I can. The settings pages are so cluttered as is.

But I do agree that having it at >15s is awkward and tough to find. Let me think more on it.

PR #1644, opened 2026-06-29
New user preference to disable Dolby Vision

whitingj:
The PR adds a user preference to control dolby vision support in device profiles. This gives the user the ability to turn off problematic dolby vision media without relying on KnownDefects.kt to handle all permutations. 

Background:
My household uses FireTV for streaming our media. I have several different models across multiple generations connected to different brands of TVs. It has been a very frustrating experience trying to get Dolby Vision media to play on those devices. I frequently get a black screen when play DV media by it various by model and TV. This seems to be a very [common issue](https://github.com/jellyfin/jellyfin-androidtv/issues/4021) that continues to cause people problems. I tried various fixes including filtering DV codecs with mixed success (it worked on some FireTV / TV combos but not others). Since the problem varies depending on the TV and media, I thought it best to give the user a preference to disable DV compatibility and let the server transcode. 

Feature Changes:
* **AppPreference.kt**: Added the new user preference. The preference is labeled "Disable Dolby Vision" and defaults to "Off". It provides two other options "Disable All Dolby Vision" and "Disable Dolby Vision with HDR10". The first removes all Dolby Vision compatibility, when the second only removes compatibility with hybrid Dolby Vision/HDR10 media. (See Screen Shoot)
* **DeviceProfileService.kt**: Glues up the new user preference with DeviceProfileUtils.kt.
* **DeviceProfileUtils.kt**: In HEVC and AV1 unsupported buildSet it will apply the user preference and mark the device as incompatible.
* **WholphinDataStore.proto**: Updates to the protobuf to store the new user preference.
* **strings.xml**: The new strings I introduced for the prefence.

## Description

Adding a new user preference to disable Dolby Vision when sending device compatibility information to the Jellyfin server.

### Related issues

This is a common problem across the Jellyfin ecosystem, but there isn't a specific issue for Wholphin. My experience is the issue is FireTV (model and generation) + TV dependent + media dependent. I tested videos from https://kodi.wiki/view/Samples with the same DOVIWithHDR10 flag that worked while other movies I have with the same flag didn't work. It is a very frustrating user experience and the current work around of setting the bitrate really low to force a transcode is sub-optimal IMO as it lowers the video quality. This allows the user to maintain full quality while disabling DV.

https://github.com/jellyfin/jellyfin-androidtv/issues/4021
https://github.com/jellyfin/jellyfin-androidtv/issues/4892
https://github.com/nova-video-player/aos-AVP/issues/1503
https://www.reddit.com/r/jellyfin/comments/12jqoxh/dolby_vision_video_not_working_on_fire_stick_4k/

### Testing

FireTV 4K (gen 1) + Vizio
FireTV 4K Max (gen 1) + LG
FireTV 4K Max (gen 2) + Vizio
FireTV (gen 1) + Olevia

## Screenshots

Preference-Choices
Prefence

## AI or LLM usage

I used Claude Code to help guide me through the unfamiliar repo and to prototype solutions. The current code in the PR is my own.


damontecres (changes requested):
This PR has build script changes unrelated to the suggested feature. I will review this PR once that is fixed.

whitingj:
Removed the changes to the build and github actions. Let me know if I need to make any other changes before getting a review. Thanks.

xAphex:
At this point there needs to be a Dolby Vision section that includes all Dolby vision settings, or else it clutters settings.

image


another thing, when you select disable dolby vision with hdr10, does it direct play HDR10 or transcodes it?

app/src/main/java/com/github/damontecres/wholphin/preferences/AppPreference.kt
```diff
@@ -443,6 +443,20 @@ sealed interface AppPreference<Pref, T> {
                 summary = R.string.force_dovi_profile_7_summary,
             )
 
+        val DoviDeviceCompatibilityPref =
```
damontecres, line 457: I'd like to make this an experimental setting, see #1624 for an example.

app/src/main/java/com/github/damontecres/wholphin/util/profile/DeviceProfileUtils.kt
```diff
@@ -448,6 +450,20 @@ fun createDeviceProfile(
                     if (!mediaTest.supportsAV1HDR10()) add(VideoRangeType.HDR10.serialName)
                 }
             }
+
+            // Check the user preferences for DoVi compatibility mode
+            when (doviDeviceCompatibilityMode) {
+                DoviDeviceCompatibilityMode.DISABLE_DOVI_ALL -> {
+                    add(VideoRangeType.DOVI.serialName)
+                    add(VideoRangeType.DOVI_WITH_HDR10.serialName)
```
damontecres, line 489: Does `DOVI_WITH_HLG` and `DOVI_WITH_SDR` need to be considered here?

app/src/main/java/com/github/damontecres/wholphin/util/profile/DeviceProfileUtils.kt
```diff
@@ -448,6 +450,20 @@ fun createDeviceProfile(
...
+                    if (jellyfinTenEleven) add("DOVIWithHDR10Plus")
+                }
+                DoviDeviceCompatibilityMode.DISABLE_DOVI_WITH_HDR10 -> {
+                    add(VideoRangeType.DOVI_WITH_HDR10.serialName)
+                    if (jellyfinTenEleven) add("DOVIWithHDR10Plus")
+                }
+                else -> {}
```
damontecres: Use exhaustive enum list instead of `else` so potential future changes force a compiler error.

app/src/main/java/com/github/damontecres/wholphin/util/profile/DeviceProfileUtils.kt
```diff
@@ -467,11 +483,30 @@ fun createDeviceProfile(
...
+                    }
+                }
+                DoviDeviceCompatibilityMode.DISABLE_DOVI_WITH_HDR10 -> {
+                    add(VideoRangeType.DOVI_WITH_HDR10.serialName)
+                    if (jellyfinTenEleven) {
+                        add("DOVIWithHDR10Plus")
+                        add("DOVIWithELHDR10Plus")
                     }
                 }
+                else -> {}
```
damontecres: Same as above

app/src/main/res/values/strings.xml
```diff
@@ -466,6 +466,17 @@
...
     <string name="force_dovi_profile_7_summary">Ignores device compatibility checks</string>
+    <string name="dovi_device_compatibility_mode">Disable Dolby Vision</string>
+    <string-array name="dovi_device_compatibility_modes">
+        <item>Off</item>
+        <item>Disable All Dolby Vision</item>
+        <item>Disable Dolby Vision with HDR10</item>
+    </string-array>
+    <string-array name="dovi_device_compatibility_mode_descriptions">
+        <item>Allow Dolby Vision media</item>
+        <item>Force transcoding by reporting the device as incompatible with all Dolby Vision media</item>
```
damontecres:
```suggestion
        <item>Force transcoding all Dolby Vision media</item>
```

app/src/main/res/values/strings.xml
```diff
@@ -466,6 +466,17 @@
...
+        <item>Force transcoding by reporting the device as incompatible with Dolby Vision and HDR10/HDR10+ hybrid media</item>
```
damontecres:
```suggestion
        <item>Force transcoding Dolby Vision with HDR10/HDR10+ hybrid media</item>
```

damontecres (changes requested):
I'll admit that I'm not super familiar with Dolby Vision and all of the peculiarities with it, but I do agree that it seems accurately detecting support is difficult so this seems useful.

Thanks for the contribution!

app/src/main/java/com/github/damontecres/wholphin/util/profile/DeviceProfileUtils.kt
whitingj: Thanks for pointing out these formats. Since the preference is disable all dolby vision, I think we should it should report the device as being incompatible with these formats (even through I haven't had a problem with it). The less restrictive option would continue to only disable the dolby vision with hdr10(+) formats.

damontecres: I think reporting all of them as incompatible makes sense

app/src/main/java/com/github/damontecres/wholphin/preferences/AppPreference.kt
whitingj: Ok I like the idea of it being an experimental setting as most people (hopefully) won't have to deal with this. I'll make it an experimental setting and not combine it with the other dolby vision setting.

xAphex:
is there a way to obtain a build with the changes here so i can test them?
i have some media that refuses to play because it forces DV.
thanks

whitingj:
> is there a way to obtain a build with the changes here so i can test them? i have some media that refuses to play because it forces DV. thanks

To test you'd have to compile and install via ADB. If that is something you are comfortable doing, I'm happy to help walk you through that. Otherwise you'll have to wait for it to land in the app. I was traveling and just got back. My plan is to get this PR ready to go today.

whitingj:
Ok I've addressed all of the comments and moved the setting to the experimental settings rather than combining it with the other Dolby Vision settings. Please review and let me know if there are anything else I need to change. Thanks.

app/src/main/java/com/github/damontecres/wholphin/services/DeviceProfileService.kt
```diff
@@ -62,8 +62,9 @@ class DeviceProfileService
                                 assDirectPlay = newConfig.overrides.assPlaybackMode != AssPlaybackMode.ASS_TRANSCODE,
                                 pgsDirectPlay = newConfig.overrides.directPlayPgs,
                                 dolbyVisionELDirectPlay = newConfig.overrides.directPlayDolbyVisionEL,
+                                doviDeviceCompatibilityMode = newConfig.overrides.doviDeviceCompatibilityMode,
                                 decodeAv1 = prefs.overrides.decodeAv1,
-                                preferAc3ForSurround = appPrefs.experimentalPreferences.enabled { preferAc3Surround },
+                                preferAc3ForSurround = newConfig.experimental.enabled { preferAc3Surround },
```
whitingj, line 69: I changed this to use newConfig so it is consistent with how all the other ones are done is this block.

whitingj:
I would love to get a review on these changes.

app/src/main/java/com/github/damontecres/wholphin/preferences/AppPreference.kt
damontecres: Can you move this to `ExperimentalPreference.kt` with the other experimental settings?

damontecres (changes requested):
Just need to move some code, otherwise looks good to me!

Edit: Please also run pre-commit to fix up any style issues.

alexkahler:
I would also flag potential UX issues with https://github.com/damontecres/Wholphin/pull/1876 "Force DV" settings and this PRs "Disable DV" settings. Both are mutually exclusive so I suggest ensuring that a user can only select 1 option from e.g., a list.

@xAphex comment in https://github.com/damontecres/Wholphin/pull/1644#issuecomment-4938101162 is a good suggestion.

damontecres:
> I would also flag potential UX issues with #1876 "Force DV" settings and this PRs "Disable DV" settings. Both are mutually exclusive so I suggest ensuring that a user can only select 1 option from e.g., a list.
> 
> @xAphex comment in [#1644 (comment)](https://github.com/damontecres/Wholphin/pull/1644#issuecomment-4938101162) is a good suggestion.

This is a good point. For now, I think neither PR should worry about the other and I'll merge them independently. Then I'll do a follow up PR to organize the settings.

xAphex:
Official jellyfin for android tv added an option to force enable or disable common HDR formats in v20 beta:
https://github.com/jellyfin/jellyfin-androidtv/pull/5743

app/src/main/java/com/github/damontecres/wholphin/preferences/AppPreference.kt
whitingj: moved.

whitingj:
Ok this PR is ready for re-review (hopefully the last 🤞🏻). I've moved the preference to`ExperimentalPreference.kt` and ran the pre-commit on all files to make sure the formatting was good. Thanks for looking at this and let me know if I need to change anything else.

whitingj:
I hate to mention this because we are almost at the finish line, but if you prefer Wholphin to copy what the official Jellyfin client is doing allowing Wholphin to enable, disable, or auto each dolby / hdr10 format, I can scratch my changes and port that over.

https://github.com/jellyfin/jellyfin-androidtv/pull/5743

If we went with this approach I would create a new dolby vision / hdr menu settings and the new switches would go here as well as the direct play dolby vision setting. The settings would be app setting and not experimental settings.

whitingj:
Just following up to see if we want to merge this or take the same approach as the official jellyfin client?

PR #1650, opened 2026-06-30
Add option for trickplay preview mode when D-Pad seeking

calebvetter:
## Description
Adds a setting to allow the user to choose between the current D-Pad skip functionality, and a new simple trickplay overlay.

The setting is added in Advanced, under the Playback section, called "D-Pad seek mode". The options are "Skip time" and "Trickplay previews". I left the default as "Skip time" since that's how it operates now, and nothing would change for users after this option is added unless they go and set it.

When "Trickplay previews" is enabled, it shows an simple overlay when using the D-Pad to seek (without the full overlay open). The new simple overlay is the same implementation of the components in the full overlay: seekbar, trickplay images, current playhead time, remaining time. The components were pulled from the full overlay's inline UI code into their own reusable components, now used in both the full overlay and the simple overlay.

#### Notes
I could see there being an adjustment for the overlay hide delay. It's currently hard coded to 1.5s.

This could be the default method, but it seems like preference is split as to whether people want this, or the minimalistic skipped seconds only version.

### Related issues
This is directly in response to this issue, as well as my own desire for the feature :)
https://github.com/damontecres/Wholphin/discussions/1084

It's also loosely related to this issue, in that it would solve it since it shows the playhead and remaining times.
https://github.com/damontecres/Wholphin/discussions/1122

### Testing
Built and tested on an Amazon Fire TV Stick 4K Max. I wouldn't anticipate and compatibility or performance changes since it's a simple settings/UI change involving already-used UI components.

## Screenshots
wholphin-20260630-155708
wholphin-20260630-155723
wholphin-20260630-153825

## AI or LLM usage
AI was used to generate the change, automate some testing, and provide information surrounding the change. I manually did some refactoring and cleanup to keep it clean and consistent with the rest of the code conventions.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/preferences/AppPreference.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/preferences/AppPreference.kt
@@ -661,6 +661,19 @@ sealed interface AppPreference<Pref, T> {
                 destination = Destination.Settings(PreferenceScreenOption.SKIP_SEGMENTS),
             )
 
+        val DpadSeekModePref =
+            AppChoicePreference<AppPreferences, DpadSeekMode>(
+                title = R.string.d_pad_seek_mode_title,
+                defaultValue = DpadSeekMode.SKIP_TIME,
+                getter = { it.playbackPreferences.dpadSeekMode },
+                setter = { prefs, value ->
+                    prefs.updatePlaybackPreferences { dpadSeekMode = value }
+                },
+                displayValues = R.array.dpad_seek_mode_options,
+                indexToValue = { DpadSeekMode.forNumber(it) },
+                valueToIndex = { if (it != DpadSeekMode.UNRECOGNIZED) it.number else 0 },
+            )
+
         val GlobalContentScale =
             AppChoicePreference<AppPreferences, PrefContentScale>(
                 title = R.string.global_content_scale,
@@ -1226,6 +1239,7 @@ val advancedPreferences =
                         AppPreference.CinemaMode,
                         AppPreference.GlobalContentScale,
                         AppPreference.SkipSegments,
+                        AppPreference.DpadSeekModePref,
                         AppPreference.MaxBitrate,
                         AppPreference.RefreshRateSwitching,
                         AppPreference.ResolutionSwitching,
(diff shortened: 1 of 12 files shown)


damontecres:
Thanks for the contribution @calebvetter ! Can you resolve the merge conflicts so I can review?

calebvetter:
Done! Let me know what you think and if there's anything you'd like to see different!

damontecres:
Thanks again! The code looks pretty solid. However, I'm not sure it's the right behavior.

This shows the trickplay _and_ seeks simultaneously, but I think the desired behavior would be how the full UI seek bar works: trickplay shows immediately, but there is a slight delay before seeking.

What do you (and others) think?

ktm0589-design:
`I think the desired behavior would be how the full UI seek bar works: trickplay shows immediately, but there is a slight delay before seeking.`

This!

calebvetter:
Ahh, yes that does sound better. My mind was stuck closer to the current dpad seek functionality when making this, haha.

Updated to use the same seek bar state update as the full UI, which has the built in 750ms debounce. Added a helper function in the playback key handler so all seek calls get funneled correctly based on user prefs, handled by the playback page in the callback function.

ScottESanDiego:
I'm not personally a fan of the current seekbar behavior, but +1 to consistency.

ktm0589-design:
Maybe a setting to set the delay time?

If this gets accepted, when will the android TV app gets the update? Does that take long usually?

calebvetter:
> Maybe a setting to set the delay time?

I could see that, but it would be outside the scope of this PR since the delay time is directly tied to the full UI's seekbar delay time (as it should be, I think). So it would probably be something for a new PR if there's enough interest in that.

> If this gets accepted, when will the android TV app gets the update? Does that take long usually?

It looks like they publish a version every couple weeks or so, as needed, which would include any features/fixes that have been merged up to that point.

damontecres:
I pushed a few changes. I prefer to separate the functions that get state from ones that perform actions, so I split the new `onDpadSeek` & `onEnterHiddenControls` functions.

I also added animation for the seek bar to slide in & out.

damontecres (approved):
Looks great! Thanks!

damontecres:
> If this gets accepted, when will the android TV app gets the update? Does that take long usually?

Not a hard and fast rule, but I've been pushing a new release about every two weeks. I'm planning to release early next week.

calebvetter:
> I pushed a few changes. I prefer to separate the functions that get state from ones that perform actions, so I split the new `onDpadSeek` & `onEnterHiddenControls` functions.
> 
> I also added animation for the seek bar to slide in & out.

Sweet! I'll keep this in mind if I work on anything else! :)

Issue #1656, opened 2026-07-02
Playing a song from Search results fails with message: "Unsupported item type: Audio"


StefanBS:
### Description

Getting this error on Android TV. Tested on current main branch, commit 06d7c8502737068f540b5240d8b4634d4022ef99.

When following the reproduction steps I get a "Unsupported item type: Audio" error message and no audio plays.

### Reproduction steps

Search -> Type the song name in the search box -> Select a result from the Song section -> "Unsupported item type: Audio"

### Experimental settings

No

### App Version

main - commit 06d7c8502737068f540b5240d8b4634d4022ef99

### Server Version

10.11.11

### Device

Sony Bravia 4K VH21

### Logs

```bash

```

StefanBS:
Searching for an open issue I found this: https://github.com/damontecres/Wholphin/issues/1573. That is a similar issue that has already been fixed by [this PR](https://github.com/damontecres/Wholphin/pull/1623) mentioned in the comments (I confirmed that issue 1573 is fixed on my local build from main). That PR fixed issue 1573 by removing access to that path from the menu.

This is a different case. In this case I think we have 2 options to fix this:
Option 1: play on click instead of navigating
Option 2: navigate to album view with cursor on song, don't play

I have full albums in my library and I would expect song search results to behave like option 2. Option 2 is how TV episode search behave in Wholphin, and also how song search on Spotify TV app works, but this is a bigger change of course.

PR #1661, opened 2026-07-03
Selecting a song on the search page navigates to a song's album with the song focused

StefanBS:
## Description
Selecting a song from search now resolves to its album's detail page, scrolled to and focusing that song, instead of sending you to an "unsupported item type" page. Implemented in the shared BaseItem.destination() rather than in search specifically, so any caller that navigates to a song gets the same behavior.

A song with no albumId plays directly while remaining in the search results page: NavigationManager.navigateTo() intercepts a bare AUDIO destination and hands it to a new MusicService.playSong(). Handling this in the single navigation funnel means every caller gets it for free with no loading-screen flash.

Adds BaseItemDestinationTest covering the new destination() logic.

### Related issues
https://github.com/damontecres/Wholphin/issues/1656

### Testing
Device: Sony Bravia 4K VH21

#### Tests on a search results page, song with albumId:
1. Select a song sends you to album page with cursor on song. Another press plays it
2. Back button brings you back to search results.

Tested with both songs at the start of an album and songs deep into a 100+ songs album.

#### Tests on a search results page, song with no albumId:
1. Select a songs plays the song while staying in the search results page. No loading screen.

## AI or LLM usage


app/src/main/java/com/github/damontecres/wholphin/services/NavigationManager.kt
```diff
@@ -13,13 +14,20 @@ import javax.inject.Singleton
 @Singleton
 class NavigationManager
     @Inject
-    constructor() {
+    constructor(
+        private val musicService: MusicService,
```
damontecres: The `NavigationManager` should not have any dependencies.

Calling `musicService.playSong` should be performed by a `ViewModel` instead

app/src/main/java/com/github/damontecres/wholphin/services/MusicService.kt
```diff
@@ -168,6 +170,27 @@ class MusicService
...
             }
 
+        /**
+         * Fetch a song by id and play it in place, without navigating.
+         *
+         * Used for album-less songs, which have no page of their own (see [NavigationManager.navigateTo]).
+         * Fire-and-forget on [defaultScope] so the caller stays on the current screen.
+         */
+        fun playSong(itemId: UUID) {
+            defaultScope.launch {
```
damontecres: Make this `suspend` instead and have `ViewModel`s handle the scope

app/src/main/java/com/github/damontecres/wholphin/ui/detail/music/AlbumDetailsPage.kt
```diff
@@ -357,25 +369,34 @@ fun AlbumDetailsPage(
...
-                }
-                viewModel.updateBackDrop()
-            }
+            RequestOrRestoreFocus(
+                if (position.row == SONG_ROW) songFocusRequester else focusRequesters.getOrNull(position.row),
+            )
             val backHandlerActive by remember {
                 derivedStateOf {
-                    listState.firstVisibleItemIndex > itemsBefore
+                    // Deep-linked opens already start scrolled to the song, so let back pop the
```
damontecres: I don't think this behavior would be what the user expects since it changes based on how the user opened the page. The back button behavior should be consistent and predictable.

app/src/main/java/com/github/damontecres/wholphin/ui/main/SearchPage.kt
```diff
@@ -435,7 +435,7 @@ fun SearchPage(
     LaunchedEffect(Unit) {
         focusRequesters.getOrNull(position.row)?.tryRequestFocus()
     }
-    val onClickItem = { index: Int, item: BaseItem ->
+    val onClickItem = { _: Int, item: BaseItem ->
```
damontecres, line 562: Your LLM is leaking

app/src/main/java/com/github/damontecres/wholphin/ui/detail/music/AlbumDetailsPage.kt
```diff
@@ -357,25 +369,34 @@ fun AlbumDetailsPage(
...
             val album = state.album!!
+            val itemsBefore = 2
+            val initialSongIndex = state.initialSongIndex
+            val hasInitialSong = initialSongIndex >= 0
+            var position by rememberPosition(
+                row = if (hasInitialSong) SONG_ROW else 0,
+                column = initialSongIndex.coerceAtLeast(0),
+            )
+
+            LaunchedEffect(Unit) { viewModel.updateBackDrop() }
```
damontecres: Not quite sure what all of this was moved around

damontecres (changes requested):
Thanks for the contribution! However, this PR needs some work.

How common are songs without albums? I think it would be better to open them to a placeholder page instead of playing directly. I think a user would not expect it to play immediately and it would disrupt the music queue they are building.

StefanBS:
> Thanks for the contribution! However, this PR needs some work.
> 
> How common are songs without albums? I think it would be better to open them to a placeholder page instead of playing directly. I think a user would not expect it to play immediately and it would disrupt the music queue they are building.

You are right. I implemented a placeholder screen for this. It looks the same as the Album page's header area with the shuffle button hidden. Everything else below the header has no analogue for a single song.
 
Screenshot_20260726_131220

app/src/main/java/com/github/damontecres/wholphin/services/NavigationManager.kt
StefanBS: This is no longer touched by this PR after we implemented the Song screen

app/src/main/java/com/github/damontecres/wholphin/services/MusicService.kt
StefanBS: This is no longer touched by this PR after we implemented the Song screen

app/src/main/java/com/github/damontecres/wholphin/ui/detail/music/AlbumDetailsPage.kt
StefanBS: Reverted

app/src/main/java/com/github/damontecres/wholphin/ui/main/SearchPage.kt
StefanBS: Not relevant anymore. This exact change was introduced to main in ed97fc30

app/src/main/java/com/github/damontecres/wholphin/ui/detail/music/AlbumDetailsPage.kt
StefanBS: L374 and L380 have been removed now. L372 has moved up because it's now used by listState

damontecres (approved):
This looks great! Thanks for the contribution!

I pushed a small change to apply the same behavior on a Music library's Songs tab too.

PR #1654, opened 2026-07-01
Improve track selection logic

damontecres:
## Description
Updates the logic for selecting tracks which should fix several edge cases

The new logic groups track types together and then applying the user's selection to the actual tracks in the file. For example, if the user choosing the audio track which the server reports as absolute `index=2` (eg the _third_ track of any type in the file) and audio `index=1` (eg the _second audio_ track in the file), then the track selection logic will use the _second audio_ track in the direct played file (after accounting for id sorting).

Some other changes:
- If direct playing and an unsupported track is selected, the app will now fall back to transcoding
- If the correct audio/subtitle tracks are not found, the app will now fall back to transcoding instead of blocking any track change attempts

### TODO
Not necessarily in this PR, but some things need further investigation:
- Handling external audio tracks
- Why does ExoPlayer order the tracks differently sometimes, see [example here](https://github.com/damontecres/Wholphin/issues/791#issuecomment-4836470813) where `0:3` is _after_ `0:4`, but the `MediaSourceInfo` lists them `0:3` then `0:4`

### Related issues
Maybe fixes #791
Should fix https://github.com/damontecres/Wholphin/issues/1005#issuecomment-4085440175
Related to the parent issue, #1475
Replaces #845

### Testing
Unit tests mostly, some basic tests on emulator

Needs more "real world" testing

#### Try it out?

You can try out this build by following the [instructions here](https://github.com/damontecres/Wholphin/releases/tag/develop-track-selection). If you do, please report your results in this PR!

## Screenshots
N/A

## AI or LLM usage
None

PR #1684, opened 2026-07-10
Fix ExoPlayer accessed on wrong thread when changing tracks

kirill578:
## Description
`changeStreamsDirectPlay` was reading `player.currentTracks` on the IO dispatcher. ExoPlayer requires player access on the main thread, so switching audio/subtitle tracks during direct play could crash with `IllegalStateException: Player is accessed on the wrong thread`.

This reads the tracks via `onMain` first (same pattern already used elsewhere in that function), then uses that value for `checkForSupport`.

### Related issues
No existing issue. Reproduced while changing subtitle tracks during direct play; logcat showed:
`Player is accessed on the wrong thread` / `Expected thread: 'main'` / `Current thread: 'DefaultDispatcher-worker-…'` from `PlaybackViewModel.changeStreamsDirectPlay`.

### Testing
- Reproduced the crash on device before the fix when changing tracks during direct play
- After the fix, changing audio/subtitle tracks during direct play no longer crashes

## Screenshots
N/A

## AI or LLM usage
Cursor/Claude helped locate the crash from logcat and apply the one-line fix. I reviewed the change and can maintain it.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackViewModel.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackViewModel.kt
@@ -912,11 +912,12 @@ class PlaybackViewModel
                         } else {
                             _state.update { it.copy(currentItemPlayback = itemPlayback) }
                         }
+                        val currentTracks = onMain { player.currentTracks }
                         _state.update {
                             it.copy(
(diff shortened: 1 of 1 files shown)


kirill578:
disregard looks like another pr just fixed it https://github.com/damontecres/Wholphin/pull/1654

PR #1687, opened 2026-07-11
Fix subtitle outline width not applying in release builds

### LINKED ISSUE #1681: Subtitle edge size doesn't apply
### Description

Adjusting the subtitle edge doesn't change in ExoPlayer

### Reproduction steps

1. Open Settings->Subtitle style
2. Change "Edge size"
3. See preview doesn't change & playback doesn't change

### Experimental settings

No

### App Version

v1.0.3

### Server Version

10.11.11

### Device

SHield 2019 pro

### Logs

```bash

```

damontecres:
## Description
Fix applying the subtitle outline width in ExoPlayer for release builds

The outline width field isn't public, so reflection is used to bypass and modify it. Since proguard/r8 renames all of the classes and fields, this meant the field was no longer found. The fix is to have proguard preserve the necessary field names.

### Related issues
Fixes #1681

### Testing
Emulator w/ release build

## Screenshots
N/a

## AI or LLM usage
None

### DIFF
--- a/app/proguard-rules.pro
+++ b/app/proguard-rules.pro
@@ -11,3 +11,14 @@
 
 # TODO investigate using smaller scope
 -keep class com.google.common.cache.** { *; }
+
+# Media3SubtitleOverride uses reflection to access/modify these fields
+-keep class androidx.media3.ui.SubtitleView {
+  private androidx.media3.ui.SubtitleView$Output output;
+}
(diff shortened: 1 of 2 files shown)


damontecres:
Note: the only other place reflection is used in Wholphin is in `WholphinRenderersFactory` to construct the `dav1d` renderer if needed. This is covered by upstream proguard configuration from media3.

PR #1720, opened 2026-07-18
Route multichannel FLAC to the FFmpeg decoder instead of the AOSP decoder (SHIELD)

### LINKED ISSUE #1719: 5.1/multichannel FLAC fails to direct play on ExoPlayer (SHIELD)
### Description

Files whose audio is FLAC 5.1 will play until a certain point then hang or fall back to transcoding on my nvidia shield. The same files will play fine with their included FLAC 2.0 audio tracks.

### Experimental settings

No

### Player backend

ExoPlayer (default)

### Media info

mkv HEVC Main 10 audio FLAC 5.1 (6-channel, 48 kHz) w/ secondary FLAC 2.0 audio

### App Version

v1.0.3

### Server Version

10.11.11

### Device

NVIDIA SHIELD Android TV (2019, "sif"), Android 11

### Logs

```bash
SoftFlacDecoder: onQueueFilled: initially configuring decoder: 48000 Hz, 6 channels
FLACDecoder: decodeOneFrame: no streaminfo metadata block   (repeated)
MediaCodecRenderer$DecoderInitializationException: Decoder init failed: [-49999], Format(3, FLAC 5.1, video/x-matroska, audio/flac, null, -1, en, [-1, -1, -1.0, null], [6, 48000])
PlaybackViewModel$onPlayerError: Playback error during DirectPlay, falling back to transcoding
```

sudolulo:
The AOSP software FLAC decoder (c2.android.flac / OMX.google.flac) advertises multichannel support on the 2019 nvidia shield "sif" but 5.1 FLAC can't Direct Play and stalls or falls back to transcoding. 

## Description

This reports multichannel FLAC as unsupported in a MediaCodecAudioRenderer subclass so the track selector routes it to the bundled FFmpeg extension decoder, but only when the device is a 2019 nvidia shield non pro (sif).

### Related issues

Fixes #1719

play a file with FLAC 5.1 audio on an nvidia shield and wait

### Testing

Tested on my NVIDIA SHIELD Android TV 2019. I have been running this fix and it allows previously broken files to play across multiple series compared to the unmodified client. It does not alter the clients ability to play any other audio types or cause them to fall back to transcoding.

Files with multichannel FLAC audio fall back to the ffmpeg decoder while FLAC 2.0 files continue to use the default decoder. Nothing else changes.

## Screenshots

no ui changes

## AI or LLM usage

Diagnosis and the initial patch were assisted with AI.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/services/PlayerFactory.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/services/PlayerFactory.kt
@@ -39,6 +39,7 @@ import com.github.damontecres.wholphin.preferences.PlayerBackend
 import com.github.damontecres.wholphin.preferences.get
 import com.github.damontecres.wholphin.services.hilt.AuthOkHttpClient
 import com.github.damontecres.wholphin.util.WholphinDispatchers
+import com.github.damontecres.wholphin.util.profile.KnownDefects
 import dagger.hilt.android.qualifiers.ApplicationContext
 import io.github.peerless2012.ass.media.AssHandler
 import io.github.peerless2012.ass.media.factory.AssRenderersFactory
@@ -336,11 +337,13 @@ class WholphinRenderersFactory(
             eventListener,
             out,
         )
-        // Replace the platform audio renderer with one that reports multichannel FLAC as
-        // unsupported, so the track selector routes it to the FFmpeg decoder instead. The AOSP
-        // software FLAC decoder (OMX.google.flac / c2.android.flac) claims to support it but fails
-        // at runtime with "no streaminfo metadata block"; stereo FLAC and vendor decoders are left
-        // alone. Via supportsFormat -- that, not getDecoderInfos, is what track selection consults.
+        // Devices in KnownDefects.multichannelFlacBug can't decode multichannel FLAC with the
+        // platform decoder. Replace the platform audio renderer with one that reports >2-channel
+        // FLAC as unsupported, so the track selector routes it to the bundled FFmpeg decoder;
+        // stereo FLAC and every other format keep the platform decoder, unchanged. Replaced in
+        // place (not appended) to preserve renderer order, and done via supportsFormat -- that,
+        // not getDecoderInfos, is what track selection consults.
+        if (!KnownDefects.multichannelFlacBug) return
         val platformIndex = out.indexOfFirst { it is MediaCodecAudioRenderer }
         if (platformIndex < 0) return
         out[platformIndex] =
@@ -357,16 +360,8 @@ class WholphinRenderersFactory(
                     mediaCodecSelector: MediaCodecSelector,
                     format: Format,
                 ): Int {
-                    if (MimeTypes.AUDIO_FLAC == format.sampleMimeType && format.channelCount > 2) {
-                        val decoders = super.getDecoderInfos(mediaCodecSelector, format, false)
-                        val aospSoftwareOnly =
-                            decoders.isNotEmpty() &&
-                                decoders.all {
-                                    it.name.startsWith("OMX.google.") || it.name.startsWith("c2.android.")
-                                }
-                        if (aospSoftwareOnly) {
-                            return RendererCapabilities.create(C.FORMAT_UNSUPPORTED_SUBTYPE)
-                        }
+                    if (format.sampleMimeType == MimeTypes.AUDIO_FLAC && format.channelCount > 2) {
+                        return RendererCapabilities.create(C.FORMAT_UNSUPPORTED_SUBTYPE)
                     }
                     return super.supportsFormat(mediaCodecSelector, format)
                 }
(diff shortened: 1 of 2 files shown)


sudolulo:
Could use further testing to be sure its an AOSP decoder issue and not nvidia shield specific but I don't have the hardware. In that case this could be narrowed in scope.

DrHurt:
This sounds like a device limitation. My TV plays 7.1 FLAC just fine.

sudolulo:
> This sounds like a device limitation. My TV plays 7.1 FLAC just fine.

do you use an nvidia shield or other android player

Does works fine mean it doesnt transcode?

damontecres:
This definitely needs more investigation and shouldn't be applied globally for all devices.

For the nvidia shield, looks like a couple people online suggest disabling Dolby Processing in the OS advanced sound settings.

sudolulo:
> This definitely needs more investigation and shouldn't be applied globally for all devices.
> 
> For the nvidia shield, looks like a couple people online suggest disabling Dolby Processing in the OS advanced sound settings.

I just tried a couple of the timestamps I've been using and disabling Dolby processing in advanced settings did not change anything. The same timestamps plays fine on my debug build with this fix. 

I am narrowing the scope of this PR back to being just the nvidia shield 2019 non pro.

DrHurt:
The problem with multichannel FLAC is not exoplayer. 

It’s usually the TV’s HDMI LPCM channel limit.

Exoplayer decodes FLAC 7.1 to LPCM 7.1 and if your TV does not support multichannel LPCM over HDMI, it’ll either need to be re-encoded to stereo PCM or re-encoded to AC3 to passthrough as multichannel. 

Few TVs support multichannel LPCM over HDMI.

sudolulo:
> The problem with multichannel FLAC is not exoplayer.
> 
> It’s usually the TV’s HDMI LPCM channel limit.
> 
> Exoplayer decodes FLAC 7.1 to LPCM 7.1 and if your TV does not support multichannel LPCM over HDMI, it’ll either need to be re-encoded to stereo PCM or re-encoded to AC3 to passthrough as multichannel.
> 
> Few TVs support multichannel LPCM over HDMI.

Would the fact that this is a decoder error mean it is upstream of the HDMI? Forgive me if I am incorrect. I am not swapping anything with exoplayer, I am swapping the renderer used from MediaCodecAudioRenderer to FfmpegAudioRenderer. Both then feed multichannel LPCM over HDMI is my understanding.

Personally I am feeding a Denon AVR-X1800H receiver that can handle LPCM 7.1 over HDMI, and am consuming multichannel audio with this fix.

damontecres:
I haven't been able to replicate the `DecoderInitializationException` from the issue on my nvidia shield 2019 pro. I still need to test other devices including a shield 2019 non-pro.

Are you able to share sample media that causes the error?

jantari:
> The problem with multichannel FLAC is not exoplayer.
> 
> It’s usually the TV’s HDMI LPCM channel limit.
> 
> Exoplayer decodes FLAC 7.1 to LPCM 7.1 and if your TV does not support multichannel LPCM over HDMI, it’ll either need to be re-encoded to stereo PCM or re-encoded to AC3 to passthrough as multichannel.
> 
> Few TVs support multichannel LPCM over HDMI.

I have a weird situation with a TCL 65C7K connected to a Onkyo TX-NR7100 via HDMI eARC. I run Wholphin directly on the TV.

When I play content in Wholphin, Dolby TrueHD 5.1 tracks get sent to the receiver as "MCH PCM 5.1". So multichannel PCM via HDMI definitely works, I guess some component just doesn't support bitstreaming TrueHD which is fine for now.

But when I attempt to play multichannel 5.1 FLAC, it gets cut down to "PCM 2.0" and the worst part is it isn't even downmixed, it's literally just the two front L/R streams playing as-is. I can tell because dialogue is almost inaudible (Center channel is missing) and effects are missing too (rear surround channels).

What's perhaps even more weird, when I play native multichannel 5.1 PCM audio it also gets cut down to PCM 2.0 (again, no downmix, extra channels just discarded).

Also when I skip the Onkyo Receiver and set playback directly to the internal speakers of the TV, I don't get the nice codec information on the receiver but I can still very clearly hear the same problem: even without the HDMI eARC and the receiver involved at all, FLAC 5.1 and PCM 5.1 get cut back to just front left and right channels being played judging from the inaudible dialogue.

I have tried every decoder setting in the official Jellyfin Android TV app and in Wholphin but nothing fixes 5.1 FLAC and 5.1 PCM *except* the experimental setting "Prefer AC3 for Surround Sound - Always transcode non-Stereo Audio to AC3/DD". With this enabled, FLAC 5.1 and PCM 5.1 tracks play in their full "fidelity" (disregarding the fact that the codec is lossy, but no more missing channels).

Is there anything I can change on the Android TV itself to better support 5.1 PCM and 5.1 FLAC playback or is it something that Wholphin has to address?

EDIT: Also, im willing to help test any settings or beta builds regarding this issue. Besides the aforementioned TCL TV and the Onkyo AVR I also have a Yamaha R-N1000A stereo receiver (2.1 only!) with HDMI ARC input I could test with.

fatSquirrel42:
Same failure on `darcy` (SHIELD TV 2017, P2897), details and log in #1719 —
there it surfaces as `Buffer too small (32768 < 33170)` in
`MediaCodecAudioRenderer` rather than a decoder init failure, but it's the same
AOSP FLAC decoder. Would it make sense to widen the device check beyond `sif` to
cover `darcy` and `mdarcy`?

Issue #950, opened 2026-02-22
[FEA] - Theme Color Customizer


crisc98:
### Description

Currently, we have the ability to switch between preset themes. It could be cool to allow users to change the specific colors of their theme individually. We could probably just expose 4 - 8 color channels and compute the remaining ones based on those choices. I have started a prototype of this idea, but I wanted to gauge interest in the idea first. My prototype adds an new Theme option called "Custom" and a new setting on the Settings screen that allows the user to open a color picker and assign a color to each of the exposed channels. I would love feedback on this idea!

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/ba818d99-960b-4dd6-8115-037110f175c8" />

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/4ef39a62-76d3-4bb4-90fc-fce5fd3896ad" />

danpergal:
I believe this is a fantastic idea that will add significant value to the app. I am certain the developer will be excited about your input to the project.
​Many thanks.

damontecres:
This was proposed a while back in https://github.com/damontecres/Wholphin/issues/91, but after I added more built-in themes, it was abandoned. I'm okay with revisiting it though!

Just some background info:

Wholphin's color schemes are generally based Google's Material Design color system, see https://m3.material.io/styles/color/roles.

The themes were all generated created using https://material-foundation.github.io/material-theme-builder/ and then slightly tweaked. There's a library for it too: https://github.com/jordond/MaterialKolor. I think this would be the best starting point to create a new theme. Also check out [`ThemePreview.kt`](https://github.com/damontecres/Wholphin/blob/main/app/src/main/java/com/github/damontecres/wholphin/ui/theme/ThemePreview.kt) if you haven't already!

As for color selection, there are sliders available too, see [`ImageFilterSliders.kt`](https://github.com/damontecres/Wholphin/blob/main/app/src/main/java/com/github/damontecres/wholphin/ui/slideshow/ImageFilterSliders.kt). But whatever is easiest is great.

Also pinging @Pentaphon & @YogiBear12 who usually have good UI ideas.

YogiBear12:
I'm generally pretty happy with the existing themes, although I don't use anything outside of the default purple.
When I was maintaining my fork, I had created three additional themes:
- A modified version of the Blue theme that swapped your primary color with Jellyfin Blue (#00A4DC)
- A Nord based theme
[...]

PR #1721, opened 2026-07-18
Add two new themes: Red & Brown

damontecres:
## Description
Adds new two new theme colors: Red & Brown

### Related issues
Related to #91, #950
Closes #1646

## Screenshots
image

## AI or LLM usage
None


Pentaphon:
@damontecres Nice additions! I especially really like brown (coffee, mocha, etc) as an option for a dark theme that is usually either blue, grey or black on Android apps! I see no weird inconsistencies or contrast issues so this is almost good to go, but I have one nitpick; the playback buttons, when selected, have a kind of sickly, yellow-hue brown? I don't mean to be gross but it's a bit too "poopy brown" rather than an earthy, natural, pleasing to the eye brown so if you can make it a tad darker and maybe reddish instead of yellowish, it would look better? In other words, just avoid that shade and go with shades that look more robust and autumnal? Maybe something like https://www.figma.com/colors/mocha/?

I hope I'm not being too nitpicky with this. The "red" theme buttons on the video player look great and have a nice brilliant tone to them but when you see them side by side with the "brown" buttons, they make the "brown" buttons look a bit anemic, lacking in color, etc. A more robust replacement to that shade of brown would really spice things up, so to speak. 

buttons

damontecres:
> if you can make it a tad darker and maybe reddish instead of yellowish, it would look better?

Yes, I agree. I've updated the color and updated the image in the PR body. I used the mocha color you suggested and I think it looks good

PR #1746, opened 2026-07-25
Correct sort order when "Date Played" is selected in TV library.

tarka:
## Description
The TV library page offers the option to sort by date played, however when selected it reverts to the default (alphabetical).  This appears to be due to the use of `ItemSortBy.DATE_PLAYED` rather than the correct `ItemSortBy.SERIES_DATE_PLAYED`. See [BrowseGridFragment](https://github.com/jellyfin/jellyfin-androidtv/blob/master/app/src/main/java/org/jellyfin/androidtv/ui/browsing/BrowseGridFragment.java#L169) for the usage in the official Android TV client.

### Related issues
N/A

### Testing
Emulator against personal Jellyfin server.

## Screenshots
N/A

## AI or LLM usage
None


damontecres (approved):
Thanks! Looks good to me

PR #1688, opened 2026-07-11
Parse intents to search, view, or play media

damontecres:
## Description
Adds parsing for intents to search, view, or play media based on the action and parameters. This allows for remotely opening Wholphin and starting playback.

This is considered experimental, so the syntax or behaviors may change!

See the `Intents.md` file in this PR for syntax and examples.

### Related issues
Replaces https://github.com/damontecres/Wholphin/pull/738
Closes https://github.com/damontecres/Wholphin/discussions/1095

### Testing
Emulator with the test commands in the docs

## Screenshots
N/A

## AI or LLM usage
None

PR #1756, opened 2026-07-27
Add back stack for deep link navigation

calebvetter:
## Description
When testing the new deep link feature from this PR [#1688](https://github.com/damontecres/Wholphin/pull/1688)
I noticed that the back button would close the app. As in, all routes are cleared and only the playback is in the nav stack.

I figured a more desirable behavior would be to create a nav stack, always starting with the home screen, then also inserting the media item's details page, if available.

To achieve this, I made the return Target.destinations a list of destinations, and updated the other two uses in IntentService.

Open to implementation tweaks, or scrap it if this isn't how it should work, haha.

### Testing

Tested deep links, which produced the following nav stacks:
- Movie view
  - Home → Movie details
- Movie playback (with and without position)
  - Home → Movie details → Playback
- Series view
  - Home → Series details
- Season view
  - Home → Season details
- Episode view
  - Home → Episode details
- Episode playback (with and without position)
  - Home → Episode details → Playback
- Trailer/extras playback (no details page available)
  - Home → Playback
- TV channel view
  - Home → Playback
- TV channel playback
  - Home → Playback

## AI or LLM usage
AI was used to generate the initial code, debug and test, then manual refactoring and cleanup was done.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
@@ -449,7 +449,13 @@ class MainActivityViewModel
                                 val current = serverRepository.current.value
                                 if (current != null) {
                                     Timber.i("Received valid intent, switching to AppContent")
-                                    navigationManager.replace(result.destination)
+
+                                    navigationManager.goToHome()
+
+                                    result.destinations.forEach { destination ->
+                                        navigationManager.navigateTo(destination)
+                                    }
+
                                     setupNavigationManager.navigateTo(
                                         SetupDestination.AppContent(current),
                                     )
--- a/app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
@@ -450,11 +450,12 @@ class MainActivityViewModel
                                 if (current != null) {
                                     Timber.i("Received valid intent, switching to AppContent")
 
-                                    navigationManager.goToHome()
-
-                                    result.destinations.forEach { destination ->
-                                        navigationManager.navigateTo(destination)
+                                    if (result.addHomeToBackStack) {
+                                        navigationManager.reloadHome()
+                                    } else {
+                                        navigationManager.backStack.clear()
                                     }
+                                    navigationManager.backStack.addAll(result.destinations)
 
                                     setupNavigationManager.navigateTo(
                                         SetupDestination.AppContent(current),
--- a/app/src/main/java/com/github/damontecres/wholphin/services/IntentService.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/services/IntentService.kt
@@ -30,7 +30,7 @@ class IntentService
             val action = intent.action ?: intent.data?.host
             if (
                 intent.getStringParam("type") == null &&
-                    (action == Intent.ACTION_MAIN || action.isNullOrBlank())
+                (action == Intent.ACTION_MAIN || action.isNullOrBlank())
             ) {
                 return IntentResult.NoOp
             }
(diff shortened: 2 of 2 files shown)


damontecres:
Thanks for the contribution!

I had considered this originally, but then I figured people would be using the intents solely for automation. However, thinking about it more, it probably does make sense to build a reasonable backstack because another intent can just replace it, but having the backstack is useful for human users.

I haven't reviewed the code yet, but I will when I have time!

calebvetter:
Ah, yeah I see that. I think with the backstack though, it can be used for automation, plus "taking over" with the remote later. For example, my intended use is I have some NFC cards that each play a specific item, but if I want to back out with the remote and play something that doesn't have an NFC then there's no stack to back out to so it just quits.

Hukuma1:
Just ran into this myself. Thank you for the PR! It launches just fine, but when I grab the remote it quits me out. This is a very nice addition.

I know I mentioned this in the other PR, but could someone please add the ability to play a randomized playlist too? (Assuming Jellyfin backend has the ability.) That way you can do "play random cartoon" automation and the like.

app/src/main/java/com/github/damontecres/wholphin/services/IntentService.kt
```diff
@@ -28,7 +28,7 @@ class IntentService
             Timber.v("Intent extras: %s", intent.extras)
             Timber.v("Intent data: %s", intent.data)
             if (intent.getStringParam("type") != null) {
-                return getDestinationFromChannel(intent)?.let { IntentResult.Target(it) }
+                return getDestinationFromChannel(intent)?.let { IntentResult.Target(listOf(it)) }
```
damontecres: This destination shouldn't have `Home` added.

Maybe `IntentResult.Target` should have two params: list of destinations and a boolean for whether to add `Home` to the backstack as well.

app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
```diff
@@ -441,7 +441,13 @@ class MainActivityViewModel
                                 val current = serverRepository.current.value
                                 if (current != null) {
                                     Timber.i("Received valid intent, switching to AppContent")
-                                    navigationManager.replace(result.destination)
+
+                                    navigationManager.goToHome()
```
damontecres: I think reloading is a good idea to ensure if the user switches, their home page will definitely reload for them.
```suggestion
                                    navigationManager.reloadHome()
```

app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
```diff
@@ -441,7 +441,13 @@ class MainActivityViewModel
...
+
+                                    result.destinations.forEach { destination ->
+                                        navigationManager.navigateTo(destination)
+                                    }
```
damontecres: Add them in one go:
```suggestion
                                    navigationManager.backStack.addAll(result.destinations)
```

damontecres (changes requested):
Some changes are needed mainly because the deep link from the OS channel rows shouldn't have any extra back stack (ie no `Home`) because the app should exit by pressing Back from this type of launch.

app/src/main/java/com/github/damontecres/wholphin/services/IntentService.kt
calebvetter: Good call. Adjusted so channel entry doesn't add home, using a new bool addHomeToBackStack (default true so the other calls don't need to explicitly set it)

app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
calebvetter: Done.

app/src/main/java/com/github/damontecres/wholphin/MainActivity.kt
calebvetter: Nice. Hadn't noticed that method before. Updated.

damontecres (approved):
Looks good! Thanks!

Issue #1760, opened 2026-07-28
Music artist page "Albums" section is empty


StefanBS:
### Description

The Albums section in an artist's detail page shows empty.

<img width="1920" height="1080" alt="Image" src="https://github.com/user-attachments/assets/51d4b34d-b127-46f7-b780-37d2af7d525a" />

### Reproduction steps

1. Select a music library
2. Go to Artist tab
3. Select an artist with music albums in your library

### Experimental settings

No

### App Version

main - bf43a41405408d74acd2f2011afd049f02928422

### Server Version

10.11.11

### Device

Sony Bravia 4K VH21

### Logs

```bash

```

PR #1761, opened 2026-07-28
Fix empty album list on the artist details page

StefanBS:
## Description
The artist page queried its albums with parentId set to the artist's ID, but an artist is not an ancestor of its albums in Jellyfin's library tree. The artist item is a metadata stub under /config/metadata/artists while the albums live under the music library folder, so the query matched nothing and the Albums section rendered empty.

Query by albumArtistIds instead, which gives the section what it wants: the albums credited to that artist. That needs recursive, since with no parentId the search is otherwise scoped to the root folder's children, which are the libraries themselves.

### Related issues
https://github.com/damontecres/Wholphin/issues/1760

### Testing
Device: Sony Bravia 4K VH21

Select a music library -> Go to Artist tab -> Select an artist with music albums in your library

## Screenshots
Before:
Screenshot_20260728_172226

After:
Screenshot_20260728_172618

## AI or LLM usage

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/detail/music/ArtistDetailsPage.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/detail/music/ArtistDetailsPage.kt
@@ -154,7 +154,9 @@ class ArtistViewModel
                         async {
                             val request =
                                 GetItemsRequest(
-                                    parentId = itemId,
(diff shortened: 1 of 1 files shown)


damontecres:
Hey! Thanks for the contribution, I ended up going a bit further with more fixes, including the changes in this PR, in #1762. So closing this PR in favor of that one.

Issue #1774, opened 2026-07-31
Alphabet buttons scroll bar in a music library Artist tab jumps to the end of the grid


StefanBS:
### Description

When clicking on the alphabet buttons scroll bar inside a music library's Artist tab the grid scrolls all the way to the bottom

### Reproduction steps

1. Music library
2. Artist tab
3. Select a letter from the alphabet button scroll bar

Expected: jump to that letter in the grid
Actual behaviour: jumps to the botton of the grid

Note: on a smaller library it could maybe not jump to the bottom but it would still jump farther than it should

### Experimental settings

Yes

### App Version

v1.0.4

### Server Version

10.11.11

### Device

Sony Bravia 4K VH21

### Logs

```bash

```

PR #1775, opened 2026-07-31
Fix the alphabet buttons scroll bar on the Artists tab

StefanBS:
## Description
positionOfLetter counted through /Items no matter which endpoint the grid was actually paging. A music library's Artists tab pages /Artists, so the index it returned had counted every album and song before the letter - an order of magnitude more rows than the artist grid holds - and every letter jumped to the last row.

Count through the same endpoint createPager builds its pager from: /Artists for the ARTIST override, /Items for NONE. The artist request is now built once in createGetArtistsRequest and shared by both, so the two cannot drift apart.

GetPersonsRequest has neither nameLessThan nor startIndex, so /Persons cannot be counted up to a letter at all; the PERSON override returns null, which the call sites already map to the "no jump" sentinel. That path is unreachable today - the only PERSON tab sorts by ItemSortBy.DEFAULT, so showLetterButtons is never true there and AlphabetButtons is never composed.

### Related issues
https://github.com/damontecres/Wholphin/issues/1774

### Testing
Select a music library -> Go to Artist tab -> Select a letter on the right side alphabet scroll bar

## AI or LLM usage

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/components/CollectionFolderView.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/components/CollectionFolderView.kt
@@ -378,7 +378,7 @@ class CollectionFolderViewModel
                     newPager
                 }
 
-                GetItemsFilterOverride.ARTIST ->
+                GetItemsFilterOverride.ARTIST -> {
                     ApiRequestPager(
                         api,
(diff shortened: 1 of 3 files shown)


damontecres (approved):
Good catch! This looks good to me!

Issue #1787, opened 2026-08-03
Alphabet scroll bar highlights "T" for artists starting with "The "


StefanBS:
### Description

On a Music library's Artist tab, when the cursor is on an artist the alphabet scroll bar highlights the initial letter of the artist name. Artists whose names start with "The " are sorted in the grid by the first letter after "The ", but the alphabet scroll bar highlights "T" instead.

For example, focusing "The Beatles" highlights "T" even though the artist is sorted under "B".

Same issue happens with Studios and Genre tabs in TV and Movie libraries

### Reproduction steps

1. Music library
2. Artist tab
3. Focus an artist whose name starts with "The " (eg "The Beatles")

Expected: alphabet scroll bar highlights the letter used for sorting (eg "B")
Actual behaviour: alphabet scroll bar highlights "T"

### Experimental settings

No

### App Version

v1.0.4

### Server Version

10.11.11

### Device

Sony Bravia 4K VH21

PR #1788, opened 2026-08-03
Fix alphabet highlight for names starting with "The "

StefanBS:
## Description
The alphabet scroll bar highlighted the letter of an item's *display* name while the grid sorted by the server's article-stripped `SortName`, so "The Beatles" sat under **B** but highlighted **T**.

- **Artists:** the Artists tab never requested `SortName`, so the highlight silently fell back to `Name`. It now asks for `SlimItemFields` (which already includes `SORT_NAME`), matching the sibling `/Items` branch. The count-only query used for letter positions passes `fields = null`, since a `limit = 0` request has no use for them.
- **Genres and studios:** these grids bucketed on the display name unconditionally. The sort-name fallback is extracted as `BaseItemDto.alphabetSortName` (`sortName ?: name ?: ""`) and carried onto the `Genre` and `Studio` grid items, so all three grids now derive the letter from one value.

### Related issues
https://github.com/damontecres/Wholphin/issues/1787

### Testing
Two unit tests added:
- `CollectionFolderViewModelTest` pins that the artist pager actually requests `SORT_NAME` — the field whose absence caused the bug.
- `AlphabetSortNameTest` covers both branches of the `alphabetSortName` fallback.

Manually: select a music library → Artists tab → put the cursor on an artist starting with "The " and confirm the highlighted letter matches the grid position. Same check on the genre and studio grids.

## AI or LLM usage
I used Cursor for debugging and implementing the fix, and Claude Code for a follow-up cleanup pass over the diff.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/data/model/BaseItem.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/data/model/BaseItem.kt
@@ -54,7 +54,7 @@ data class BaseItem(
         get() = type.playable
 
     override val sortName: String
-        get() = data.sortName ?: data.name ?: ""
+        get() = data.alphabetSortName
 
     val type get() = data.type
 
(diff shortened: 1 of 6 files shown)


damontecres (approved):
Looks good to me! Thanks!

PR #1796, opened 2026-08-05
Add live TV channel button controls

### LINKED ISSUE #509: [FEA] - Support Channel Switching via Up/Down Buttons in Live TV
### Description

While watching Live TV, allowing the remote’s up/down buttons to switch channels would better match traditional TV behavior. This would make it on par with apps like Pluto TV, and channel surfing would feel more natural.

aspenrt78:
## Description
Detects Channel Up/Down buttons, plus Media Next/Previous.
Only reuses those buttons while live TV is playing but normal videos keep their existing behavior.
Downloads the Jellyfin channel list in roughly the same order as the TV guide, then wraps from the last channel back to the first.
If a channel cannot start, it tries the next one; if a live stream ends naturally, it exits playback instead of automatically “surfing.”

### Related issues
Fixes #509 

### Testing
Compiled in GitBash then stream installed to "onn. 4K streaming box" 
Android TV OS version: 14 
Security patch level: March 1st, 2026
Kernel: 5.15.196-android14-11-g94532862e2fa
Android TV OS build: UR04.260304.011.b1.15051976

## Screenshots
Not Applicable

## AI or LLM usage
Patch was built using Claude Fable 5. Double checked with ChatGPT 5.6 Sol. Installed and tested working as designed on personal streaming box.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackKeyHandler.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackKeyHandler.kt
@@ -34,6 +34,12 @@ class PlaybackKeyHandler(
     private val isDpadSeekVisible: () -> Boolean = { false },
     private val onDpadSeek: (Long) -> Unit = { },
     private val dpadSeekMode: DpadSeekMode,
+    /**
+     * Whether channel up/down (and next/previous) should change live TV channels
+     */
+    private val channelKeysEnabled: () -> Boolean = { false },
+    private val onChannelUp: () -> Unit = { },
+    private val onChannelDown: () -> Unit = { },
 ) {
     private var leftHandledByRepeat = false
     private var rightHandledByRepeat = false
@@ -81,6 +87,12 @@ class PlaybackKeyHandler(
             } else {
                 // When controller is visible, its buttons will handle pulsing
             }
+        } else if (isChannelKey(it) && channelKeysEnabled.invoke()) {
+            if (it.key == Key.ChannelUp) {
+                onChannelUp.invoke()
+            } else {
+                onChannelDown.invoke()
+            }
         } else if (isMedia(it)) {
             when (it.key) {
                 Key.MediaPlay, Key.MediaPause, Key.MediaPlayPause -> {
@@ -98,11 +110,19 @@ class PlaybackKeyHandler(
                 }
 
                 Key.MediaNext -> {
-                    if (player.isCommandAvailable(Player.COMMAND_SEEK_TO_NEXT)) player.seekToNext()
+                    if (channelKeysEnabled.invoke()) {
+                        onChannelUp.invoke()
+                    } else if (player.isCommandAvailable(Player.COMMAND_SEEK_TO_NEXT)) {
+                        player.seekToNext()
+                    }
                 }
 
                 Key.MediaPrevious -> {
-                    if (player.isCommandAvailable(Player.COMMAND_SEEK_TO_PREVIOUS)) player.seekToPrevious()
+                    if (channelKeysEnabled.invoke()) {
+                        onChannelDown.invoke()
+                    } else if (player.isCommandAvailable(Player.COMMAND_SEEK_TO_PREVIOUS)) {
+                        player.seekToPrevious()
+                    }
                 }
 
                 Key.Captions -> {
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackKeyHandler.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackKeyHandler.kt
@@ -34,6 +34,11 @@ class PlaybackKeyHandler(
     private val isDpadSeekVisible: () -> Boolean = { false },
     private val onDpadSeek: (Long) -> Unit = { },
     private val dpadSeekMode: DpadSeekMode,
+    /**
+     * Whether channel up/down (and next/previous) should change live TV channels
+     */
+    private val channelKeysEnabled: () -> Boolean = { false },
+    private val onChangeChannel: (delta: Int) -> Unit = { },
 ) {
     private var leftHandledByRepeat = false
     private var rightHandledByRepeat = false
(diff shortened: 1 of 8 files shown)


aspenrt78:
Bumping this one, it implements #509 (labeled `enhancement`).

I've been running this daily on an onn. 4K Streaming Box for about a month now with no issues. Channel up/down and media next/previous surf the channel list while live TV is playing, and normal video playback keeps its existing behavior.

One thing I could use help with: no CI has run on this branch, so there are no build/pre-commit checks on the PR. If you can approve the workflow run I'll fix anything it turns up.

Happy to rebase onto current `main` if that's useful. I've been running the change on top of v1.0.7 locally and it applies cleanly.

app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackViewModel.kt
```diff
@@ -403,6 +410,83 @@ class PlaybackViewModel
             }
         }
 
+        /**
+         * Load the live TV channel list into the playlist so that channel up/down can move through it.
+         *
+         * Uses the same sorting as the TV guide so surfing order matches what the user sees there.
+         */
+        private suspend fun loadChannelPlaylist(currentChannelId: UUID) {
```
damontecres, line 433: This function duplicates a lot of code in `LiveTvViewModel` so would be good to migrate this to a service

app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackViewModel.kt
```diff
@@ -403,6 +410,83 @@ class PlaybackViewModel
...
+                    if (!state.isLiveTv || items.size < 2) {
+                        Timber.v("Not live TV or nothing to surf, ignoring channel change")
+                        return@withLock
+                    }
+                    if (attempt >= items.size) {
+                        Timber.w("No playable channel found after %d attempts", attempt)
+                        return@withLock
+                    }
+                    val size = items.size
+                    val newIndex = ((state.playlistIndex + delta) % size + size) % size
```
damontecres: What is this?

app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackViewModel.kt
```diff
@@ -403,6 +410,83 @@ class PlaybackViewModel
...
+            if (delta == 0) return
+            viewModelScope.launchDefault {
+                playlistMutex.withLock {
+                    val state = state.value
+                    val items = state.playlist.items
+                    if (!state.isLiveTv || items.size < 2) {
+                        Timber.v("Not live TV or nothing to surf, ignoring channel change")
+                        return@withLock
+                    }
+                    if (attempt >= items.size) {
```
damontecres: I don't see the purpose for the `attempt` variable.

Why not just calculate the newIndex with `state.playlistIndex+delay` and check if its within `playlist.items.indices`?

app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackKeyHandler.kt
```diff
@@ -34,6 +34,12 @@ class PlaybackKeyHandler(
     private val isDpadSeekVisible: () -> Boolean = { false },
     private val onDpadSeek: (Long) -> Unit = { },
     private val dpadSeekMode: DpadSeekMode,
+    /**
+     * Whether channel up/down (and next/previous) should change live TV channels
+     */
+    private val channelKeysEnabled: () -> Boolean = { false },
+    private val onChannelUp: () -> Unit = { },
+    private val onChannelDown: () -> Unit = { },
```
damontecres: Maybe just use one function that takes the delta param

app/src/main/java/com/github/damontecres/wholphin/ui/playback/PlaybackViewModel.kt
```diff
@@ -1044,6 +1129,13 @@ class PlaybackViewModel
             }
 
         override fun onPlaybackStateChanged(playbackState: Int) {
+            if (playbackState == Player.STATE_ENDED && state.value.isLiveTv) {
+                // The channel list is the playlist during live TV, so don't treat a dead
+                // stream as "next up" and zap the user to another channel
+                Timber.d("Live TV stream ended")
```
damontecres, line 1137: Maybe also show a toast? The user might be confused if playback stops suddenly and they're sent back to the previous page.

damontecres (changes requested):
Thanks for the contribution! I'm sorry it's taken a while to review!

damontecres:
>One thing I could use help with: no CI has run on this branch, so there are no build/pre-commit checks on the PR. If you can approve the workflow run I'll fix anything it turns up.

You can run `pre-commit` locally, see [here](https://github.com/damontecres/Wholphin/blob/main/DEVELOPMENT.md#development-environment).

> Happy to rebase onto current main if that's useful. I've been running the change on top of v1.0.7 locally and it applies cleanly.

Definitely merge in or rebase on `main`, whichever you prefer.

aspenrt78:
Thanks for the review, no worries about the wait.

I've pushed an update that covers everything:

Pulled the channel fetching out into a new `LiveTvChannelService`. `LiveTvViewModel` uses it now too, so the request only lives in one place instead of being duplicated.

Dropped the `attempt` variable. `changeChannel` now just does `state.playlistIndex + delta` and checks it's within `items.indices`, like you suggested. Much simpler to follow. One side effect worth mentioning: surfing no longer wraps from the last channel back to the first, and it no longer skips past a channel that fails to start. Happy to add either back if you'd rather keep them, but the simpler version seems reasonable to me.

Collapsed `onChannelUp` and `onChannelDown` into a single `onChangeChannel(delta)`.

Added a toast when a live stream ends before it navigates back. Good call, being dumped back to the guide with no explanation was confusing.

Ran `pre-commit` locally and everything passes now (ktlint, xml, whitespace, the debug checks). Thanks for the pointer to the docs, I'd missed that.

Also rebased onto current `main`, so this is a single clean commit on top of `1778bdb3`.

Issue #1798, opened 2026-08-06
Audio track persistence fails for multi-audio files with identical language tags


LessRinn:
### Description

Hi! I'd like to add a core issue regarding audio track persistence and handling across TV show episodes.

**The Problem:**
1. **Multiple Tracks with the Same Language:** Media files often contain multiple audio tracks sharing the exact same language tag (e.g., multiple `rus` tracks for different voice-over studios, differentiated only by title or index). 
2. **Lack of Persistence / Resetting to Default:** When watching a TV show, Wholphin fails to properly persist and carry over the user's preferred audio track choice. Whenever a new episode starts (especially via auto-play or next-episode triggers), the app reverts to the container's default audio track instead of remembering the previously selected option.
3. **Manual Overhead:** Because of this, users are forced to manually switch their preferred audio track for **every single episode**, as the app doesn't remember or match the preferred track across files that contain multiple tracks of the same language.

**Request:**
Could you please improve audio track persistence and handling for files with multiple tracks of the same language? Once a user selects a specific track, it should ideally persist or be remembered for subsequent episodes, rather than always reverting to the file's default track.

### Experimental settings

Yes

### App Version

v,1.0.5

### Server Version

10.11.11

### Device

xiaomi mi box 2s

### Logs

```bash

```

damontecres:
The app should be persisting the chosen language and preferring that on subsequent episodes.

Can you add the media's info by using the "Send media info log to server" within Wholphin? Ad please [gather the app logs](https://github.com/damontecres/Wholphin/wiki/Gathering-app-logs).

Also, what are your settings for:
- Audio language
- Subtitle language
- Subtitle mode

damontecres:
Oh I reread your issue, you're talking about remembering the track within the same language. That would be this feature request: https://github.com/damontecres/Wholphin/discussions/1136

LessRinn:
> Oh I reread your issue, you're talking about remembering the track within the same language. That would be this feature request: [#1136](https://github.com/damontecres/Wholphin/discussions/1136)

Yes, you understood me right! I hope this feature gets some attention at some point. You are making a great app, thank you!

PR #1802, opened 2026-08-07
Improve track persistence for series & episodes

### LINKED ISSUE #1783: Subtitles: Multiple unexpected behaviours
### Description

- Default subtitle track for a series doesn’t load, doesn’t give an error, just no subtitle appears on screen.
- Switching to another subtitle while watching a series makes every following episode load that same subtitle track. Even if you manually select the default subtitle track while watching an episode, the next episode will revert to the non-default subtitle track selected earlier.
- Subtitles appear when they shouldn’t sometimes (Forced, native language scene)
- Subtitles don’t appear when they should sometimes (Forced, non-native language scene)
- Sometimes subtitles behave as expected and appear when they should. I _think_ this might happen if a non-default, non-forced subtitle track is selected after playback of an episode begins.

### Reproduction steps

- Set default subtitle track for a show. I use `Forced Only` . Let’s call the default subtitle track “Track A” 
- Play the show: Subtitles do not appear in alternate language scenes
- Change the subtitle track to a different selection until some subtitles appear. Let’s call this “Track B”
- Finish watching the episode
- Start watching the next episode: Track B subtitles will load automatically even though Track A is the default.
- Change the subtitle track to Track A
- Finish watching the episode
- Start watching the next episode: Track B subtitles will load automatically again, no matter how many times this is repeated.

Image
Image
Image
Image
Image
Image
Image

Image

Image
Image
Image

### Experimental settings

No

### App Version

1.0.3

### Server Version

10.11.8

### Device

CCwGTV HD

### Logs

```bash

```

### LINKED ISSUE #1795: Subtitles are changed only once for the entire series
### Description

I hope I can explain the problem clearly enough. The default subtitle setting works fine for the most part, but sometimes I have to briefly switch to “Full Subtitles” during playback because otherwise I can't understand what's being said. Afterward, I switch back to the usual “Forced / Default” subtitles. Problem: Since version 1.0.3, the change to “Full Subtitles” has been applied to all episodes and remains there, even after I’ve switched back. The only “solution” I’ve found so far would be to manually change all episodes back, which would be a nightmare depending on the number of episodes.

Up to and including version 1.0.2, I can briefly switch to “Full Subtitles,” switch back, the subtitles for every other episode stays at “Forced / Default”. Here's another example, in case it helps: By default, a series with no forced subtitles plays without any subtitles. If I switch to a  subtitle while playing the series, all episodes switch from having no subtitles to that subtitle or similar.

Subtitle Mode is set to Only Forced and Set subtitle track based on previous item ist enabled (Server). It doesn't matter whether I let the server choose the subtitles or use the new feature in Wholphin.

I tested it with three different anime series, and the result was the same every time.

Image

Image

Image

### Reproduction steps

1. switch subtitles
2. switch back
3. first switch will stay active for all episodes

### Experimental settings

No

### App Version

1.0.3, 1.0.4, 1.0.5

### Server Version

10.11.11

### Device

Google TV Streamer

### Logs

```bash

```

damontecres:
> [!Caution]
> This is a work in progress! There may be bugs!!

## Description
This PR overhauls how audio and subtitle track persistence works for series! Previously, when you chose an audio or subtitle track for an episode, only the language was saved and used for subsequent episodes. This meant aspects about why that choice were lost such as a forced or hearing impaired track.

This PR changes the persistence to include more details about the chosen track. Additionally, the persistence is saved at both the season and series levels.

## How it works

Selecting the track uses the first success option:
1. Episode specific - User has explicitly chosen an audio/subtitle track for this episode
2. Season specific - User has chosen the track for another episode in this season
3. Series specific - User has chosen the track for another episode in this series
4. User preferences - User has _not_ chosen a track for the episode, season, nor series, so determine track based on the preferred audio language, preferred subtitle language, and subtitle mode

Additionally, the season or series specific choices use a scoring algorithm to find the best suited track. For example, if the chosen tracks for an episode in the season had the forced flag, this will be preferred for subsequent episodes. Same for SDH/hearing impaired or title, etc. This means that the common case where a season has substantially identical tracks will use the same ones across all episodes (after choosing for one, otherwise falling back to the series or user preferences).

Finally, this also means that users that prefer SDH/hearing impaired subtitles can be prioritized now.

The previous track language choices are migrated to the new algorithm. They lack the additional context, but this makes for a seamless transition.

## UI changes

Still WIP, but the context menu for choosing audio or subtitle tracks now displays why that track was chosen

## TODO

- [ ] More testing!!
- [ ] Review/improve the scoring algorithm
- [ ] Should the scoring change based on if the source is season vs series?
- [ ] ???

### Related issues
Fixes #1795
Maybe fixes #1783
Related to #1136 & #1798
Closes #1470

### Testing
Previous unit tests for track selection all pass, plus there are a few additional unit tests

I did basic tests with the emulator, but this needs a lot more

## Screenshots
image
image
image

## AI or LLM usage
None

## Try it out

You can try this unstable development branch by following the instructions here: https://github.com/damontecres/Wholphin/releases/tag/develop-episode-track-persist

**Note: this PR contains database changes!! This means if you install this branch, you CANNOT switch to another development branch nor stable unless you uninstall the app first!! The app will not start otherwise!**


LessRinn:
It worked, thank you!

BlackstormFirst:
Hello, I’ve just switched from the official client to your version due to various bugs and design issues—but primarily because of the poor handling of audio and subtitle streams. I hope this fix meets my expectations. Thanks a lot for your efforts in this area.

Edit:

This not retain the current audio codec type for example and i am still in "default" mode.
(I managed to force the subtitle type for the entire series, but there was a misalignment when I reached an intermediate episode that didn't have the same number of languages and in the next episode, with the same number of languages, the subtitles didn't match the series' forced selection but rather the one from the previous episode.)

I just tested this branch, and it still doesn't quite meet my expectations, even though it's getting closer.
Here is how I see things...

By default:
When you access the first item, the server sends its "best indices"—determined by the server-side strategy (in v12, the suggested indices seem improved, especially regarding forced subtitles, though this remains to be verified).
Then, by default, all subsequent videos (previous/next) should rely on the user's current selection or fallback to the most appropriate option, while maintaining that selection.

"Smart mode" should behave roughly the same way but must check each item to see if the streams (audio/subtitles) match the user's preferences (this allows, for instance, reverting to the correct language and subtitles when switching from a multi-language episode to a single-language one, and then back to content in your preferred language). However, note that if the audio language matches your preferred language, the system should select the forced subtitle track—or no subtitles at all if only full subtitle tracks exist.

"Always default" mode always loads the default audio streams but still attempts to adjust the subtitles (e.g., if a default subtitle track exists but isn't marked as "forced" while the audio language matches, the subtitle track should not be loaded).

There are many possible variations, but your implementation of modes and languages ​​based on server settings certainly facilitates the integration of an audio and subtitle manager.

I should point out that all indicators and their variants must be taken into account to implement the best possible fallbacks.

PR #1828, opened 2026-08-12
Handle density configuration changes to prevent unnecessary app restarts

### LINKED ISSUE #1809: 1.0.5 Exoplayer and MPV won't play 720p or below files when Resolution matching is ON
### Description

Here's the sample: https://transfer.it/t/3XruOk5xQJwK

Oddly, this file will not play. When you press play, it just goes back to the episode detail menu after showing the loading swirl, and sometimes the swirl won't appear at all...

Update: I just did a clean install and logged the behavior for you to see here:

[upload_Wholphin_1.0.5-0-gfceb5f81_20260810005734_26e551c7ae78465b83c21ababa9d4cdd.log](https://github.com/user-attachments/files/30882549/upload_Wholphin_1.0.5-0-gfceb5f81_20260810005734_26e551c7ae78465b83c21ababa9d4cdd.log)

**Update2: It seems both Exoplayer and MPV won't play 720p or below files when Resolution matching is ON**

### Experimental settings

No

### Player backend

ExoPlayer (default)

### Media info

[upload_Wholphin_1.0.5-0-gfceb5f81_20260809183720_2456d6952c0e4917a52569dfe0156e2a.log](https://github.com/user-attachments/files/30879117/upload_Wholphin_1.0.5-0-gfceb5f81_20260809183720_2456d6952c0e4917a52569dfe0156e2a.log)

### App Version

1.0.5

### Server Version

10.11.11

### Device

Onn 4K Plus

### Logs

```bash

```

damontecres:
## Description
Add `density` to `configChanges` and handle any changes to density that occur. Thankfully, Compose handles density changes itself, so only changes where a density calculation is "remembered" needed to be updated.

Without this, some devices will recreate `MainActivity` when switching resolutions for playback. That recreation would cause playback to end immediately.

### Related issues
Fixes #1809

### Testing
Tested on Onn 4k Pro & nvidia shield

I was able to reproduce on the Onn, but not the Shield, so this might be device and/or API level specific

## Screenshots
N/A

## AI or LLM usage
None

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -51,7 +51,7 @@
             android:name=".MainActivity"
             android:exported="true"
             android:launchMode="singleTask"
-            android:configChanges="keyboard|keyboardHidden|navigation|orientation|screenSize|screenLayout|smallestScreenSize">
+            android:configChanges="keyboard|keyboardHidden|navigation|orientation|screenSize|screenLayout|smallestScreenSize|density">
             <intent-filter>
                 <action android:name="android.intent.action.MAIN" />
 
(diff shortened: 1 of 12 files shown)


Pentaphon:
Tested and confirmed to work using debug apk. Thanks so much for following up on this. I think a lot of users will appreciate this fix. Reporting to the official app as well.

Generator:
Also fixed for me.
Tested with same previous files, and no more issues starting playback

Issue #1838, opened 2026-08-14
Home music rows use tall poster cards and stretch album art


StefanBS:
### Description

On the Home screen, rows for music libraries use the same tall poster card shape as movies/TV. Album art is stretched to fit.

Expected: music libraries rows should default to squared images with the same size as they appear inside of a music library screen.

### Reproduction steps

1. Home screen
2. Music library rows show tall, stretched images

### Experimental settings

Yes

### App Version

v1.0.6

### Server Version

10.11.11

### Device

Sony Bravia 4K VH21

### Logs

```bash

```

damontecres:
How did you add the row? What type is it?

The view options for rows is configured at creation time, but there might be a edge case for music that isn't handled.

StefanBS:
I already had two music libraries in Jellyfin before Wholphin supported them. When Wholphin started supporting music they were automatically added to my Home screen as "Recently added in <library>" rows with poster-shaped cards. They also go back to poster-shaped after resetting them in settings.

StefanBS:
createDefault doesn’t special-case music, so automatic “Recently added” rows get the poster default. Fixing it there would fix new default layouts (and anyone who Resets). Existing saved music rows would stay poster until those rows are rebuilt.

damontecres:
Oh darn, `HomeSettingsService.createDefault` doesn't check any types. It should use `HomeRowPresets.getByCollectionType`. So that definitely needs to be fixed.

PR #1839, opened 2026-08-14
Use squared cards for Music rows in Home screen

StefanBS:
## Description
Home rows for music libraries used the tall poster card shape, which stretches square album art. That happened because every row always stored a concrete viewOptions value (the poster default), so the app could not tell “nobody chose a size” from “deliberately set to poster.”

Make HomeRowConfig.viewOptions nullable (null = follow the default) and resolve the default from the parent library’s collection type. Untouched music rows now use square album cards, and they will pick up future default changes.

Also:

Per-row Reset clears the choice (null) instead of writing the default value
The Wholphin Default preset’s music rows use musicDefault (128dp square); other presets are unchanged
Small refactor: resolve a home row’s parent item once when building the row display (no behavior change)
Adds HomeRowViewOptionsTest for the default-resolution and persistence behavior.

### Related issues
https://github.com/damontecres/Wholphin/issues/1838

### Testing
- Update from a v1.0.6 install with default values for all Home screen rows to this version. All TV and Movies libraries continue showing their defaults, music rows show the new default
- Update from a v1.0.6 install with custom values for Home screen TV/Movie and music rows to this version. The custom values are kept in the new version
- Revert from fix/music-how-row-square-cards with default values to v1.0.6. The Music libraries go back from squared to stretch images. Defaults don't write to config, they are computed from whatever the current default is

## Screenshots
Screenshot_20260814_224126
Music row shows the same size and ratio as inside a Music library

## AI or LLM usage
Cursor and Claude Code

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/data/model/HomeRowConfig.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/data/model/HomeRowConfig.kt
@@ -11,15 +11,21 @@ import kotlinx.serialization.SerialName
 import kotlinx.serialization.Serializable
 import kotlinx.serialization.UseSerializers
 import org.jellyfin.sdk.model.api.BaseItemKind
+import org.jellyfin.sdk.model.api.CollectionType
 import org.jellyfin.sdk.model.api.request.GetItemsRequest
 import org.jellyfin.sdk.model.serializer.UUIDSerializer
 import java.util.UUID
 
 @Serializable
 sealed interface HomeRowConfig {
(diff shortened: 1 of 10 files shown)


damontecres:
This PR is a pretty invasive change. Rows' view options defaults are configured at creation time and aren't supposed to be dynamic.

Let's discuss more in https://github.com/damontecres/Wholphin/issues/1838#issuecomment-5303695003

damontecres:
This was replaced by #1926

PR #1864, opened 2026-08-18
Update favorites page with more types

damontecres:
## Description
This PR adds many more data types as tabs to the favorites page. Additionally, only tabs for the types the user has favorites for will appear. For example, if you have no favorited music videos, that tab will not show.

Added types:
- Collections
- TV Channels
- Music albums
- Music artists
- Music videos
- Songs
- Home videos
- Photos
- Photo albums

Dev note: The tabs are implemented using keys instead of indexes which could be useful for the series overview season tabs. Plus it maybe even be adapted for the episode list to eliminate the terrible index search logic. A major difference is that the possible favorites tabs are known/static list where seasons/episodes are queried & paginated.

### Related issues
Closes #1041
Closes #1844

### Testing
Manually on emulator, plus unit tests

## Screenshots
N/A

## AI or LLM usage
None


damontecres:
Switched to draft because further changes are needed:
- [x] Need to update the home page favorite options too
- [x] Use the logic from #1857 to filter out types without corresponding library

damontecres:
Just a note on a compile error, it should be fixed, but changes to `FavoritesViewModel` will sometimes cause a runtime `NoSuchFieldError`. This happens because the incremental compilation for the inner class `TypedProvider` doesn't seem to always work well with the compose compiler.

To fix, you can clean the build and rebuild from scratch. This should only occur from incremental builds and never on a clean release build.

The stack trace looks like this:
```
 java.lang.NoSuchFieldError: No static field $stable of type I in class Lcom/github/damontecres/wholphin/ui/detail/FavoritesViewModel$TypedProvider; or its superclasses (declaration of 'com.github.damontecres.wholphin.ui.detail.FavoritesViewModel$TypedProvider' appears in /data/data/com.github.damontecres.wholphin.debug/code_cache/.overlay/base.apk/classes7.dex)
  at com.github.damontecres.wholphin.ui.detail.FavoritesPageKt.FavoritesPage$lambda$9(FavoritesPage.kt:89)
  at com.github.damontecres.wholphin.ui.detail.FavoritesPageKt$$ExternalSyntheticLambda1.invoke(D8$$SyntheticClass:0)
...
  ```

PR #1875, opened 2026-08-20
Disable "Auto play next up" when playback starts via play key

CodeCat5:
## Description
This adds a small QOL feature that I miss from Plex clients. If playback is started by using the "Play" key on your remote/keyboard then it will disable auto-play-next-up and only play that single item. Auto-play-next-up still behaves normally when playback starts using the "OK/select" button or any method besides the "Play" button.
I find this useful when I know I may fall asleep while watching since I can use the "Play" button and prevent it from potentially playing multiple episodes while I sleep.

### Testing
 - ./gradlew :app:compileDefaultDebugKotlin passes.                                                                                                                                                                                               
  - Manually verified on 2017 Nvidia Shield, 2019 Nvidia Shield Pro, and 2024 Onn 4k Pro

## Screenshots
n/a

## AI or LLM usage
Claude Sonnet 5 was used to implement this. Code reviewed by me, installed/tested on multiple devices, and works as intended.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/components/CollectionFolderView.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/components/CollectionFolderView.kt
@@ -780,7 +780,7 @@ fun CollectionFolderView(
                                     startSlideshow = true,
                                 )
                             } else {
-                                Destination.Playback(item)
+                                Destination.Playback(item, startedViaPlayButton = true)
                             }
                         viewModel.navigateTo(destination)
                     },
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/components/ItemGrid.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/components/ItemGrid.kt
@@ -228,7 +228,7 @@ fun ItemGrid(
                         position = index
                         contextMenu.showContextMenu(index, item)
                     },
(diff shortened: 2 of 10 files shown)


damontecres:
Thanks for the contributions, but I don't think this is necessary since you can disable auto play next up and/or use pass out protection.

CodeCat5:
Thanks for the response. Of course I respect your decision on whether to include this, but I'd like to make the case for reconsidering since the two suggested settings don't actually cover this case.

Both Plex and Emby have had this exact "Play vs. Select" distinction that I've submitted here for years, and it's very useful for those who have gotten accustomed to it.

  - Disabling auto-play in settings is a permanent toggle, but this is a situational need. I want it off for tonight's session, not forever. That means backing out to Settings, turning it off, resuming, then remembering to switch it back on after I wake up. That's obviously not a practical solution nor something that any end user is going to go through.
  - Passout protection is a timeout and it guesses "inactive long enough to be asleep" after the fact. Set it short and it interrupts normal watching. Set it long enough to not be annoying (mine's 2 hours) and it's already auto-played 3-4 more episodes before it's triggered. It's a good fallback, but it can't substitute for telling the app up front.

It's a minor QOL addition, but I know many former Plex and Emby users would appreciate it. It's a simple but elegant solution both competitors settled on years ago, for the same reason I'm asking for it here.

Issue #1855, opened 2026-08-17
[Feature] Option to prefer Dolby Vision over HDR10+ on hybrid DV+HDR10+ files (MediaTek TVs play them as HDR10+)


Jycreyn:
### Description

Hey, first off thanks for the app, been daily driving it for a while now.

I've been trying to figure out for weeks why Dolby Vision never kicks in on my Philips TV (MediaTek SoC) with DV profile 8.1 MKVs. Always falls back to HDR10+, direct play, no transcoding, no errors. Same thing with the official Jellyfin ATV app so it's not specific to Wholphin.

I finally isolated it by testing the same episode 3 ways:

- original MKV (DV + HDR10+) → plays as HDR10+
- same MKV but I stripped the HDR10+ metadata with hdr10plus_tool (DV untouched) → **plays as Dolby Vision**
- same stream remuxed to MP4, still with both → HDR10+ again

So the container doesn't matter and the DV signaling in the MKV is fine (dvvC mapping is there, ExoPlayer picks it up). The issue is that when a file has both DV and HDR10+ metadata, the MediaTek pipeline just goes with HDR10+ and ignores the DV RPU. Problem is basically every 4K web release these days is a DV+HDR10+ hybrid, so DV never triggers on these TVs.

Apparently Kodi 21 ran into the same thing and works around it by only exposing one of the two formats to the SoC. Probably related to androidx/media#1895 too.

Would it be possible to have something like a "prefer Dolby Vision over HDR10+" option that drops the HDR10+ SEIs when a DV decoder is being used? That would fix it for a lot of MediaTek TVs (Philips/TCL/Fire TV) without having to remux everything.

Can provide mediainfo output, logcat during playback, whatever helps.

### Experimental settings

Yes

### App Version

1.6.0

### Server Version

10.11.11

### Device

Phillips 560 oled

### Logs

```bash

```

Issue #244, opened 2025-11-18
[FEA] - Dolby Vision Compatibility Mode (Profile 7 to 8.1 Conversion)


mattheswerner:
### Description

Implement a "Dolby Vision Compatibility Mode" to ensure correct playback of files that use Dolby Vision Profile 7 (P7). This profile, common on Blu-ray rips, often includes a Full Enhancement Layer (FEL) or Minimal Enhancement Layer (MEL) that is not supported by many modern streaming devices (e.g., Nvidia Shield, Apple TV, Fire TV).

The Problem: Playing Profile 7 content on incompatible hardware results in severe color corruption (a prominent purple/green tint) because the device cannot process the enhancement layer.
The Solution: This mode should perform an on-the-fly conversion of the Profile 7 video stream into the more widely compatible Profile 8.1. It would essentially merge the enhancement data from the P7 layer with the base layer to create a valid, single-layer P8.1 stream that plays correctly on modern hardware without color issues.

Kranzes:
Could this conversion be done on the Jellyfin server side?

mattheswerner:
> Could this conversion be done on the Jellyfin server side?

Unfortunately not, the Jellyfin server does not have this function. It would also be quite complicated to set up using third-party programs on the server side. It would be best if the client could do this. Kodi is currently the only client app that has this feature; the Jellyfin for Android TV app does not support this either.

Kranzes:
Ideally though, it should be the server's job, no? The server usually handles anything to do with remuxing/transcoding/converting. I understand that it currently does not have this functionality but maybe it's possible to use dovi_tool directly, the binary, just like ffmpeg is used as a binary.

Related: https://features.jellyfin.org/posts/2468/use-dovi_tool-to-convert-dolby-vision-profile-7-to-8-1-when-necessary

mattheswerner:
> remuxing

I have one media player that supports Dolby Vision Profile 7, while another only supports Profile 8.1. If I pre-convert the media to Profile 8.1 on the server, I can no longer use the full capabilities of my Profile 7 player. It would be ideal if the client could automatically adapt the metadata to match its own playback capabilities, similar to how Kodi handles this.

Kranzes:
I'm suggesting to media to be converted on the fly by the server. With the latest jellyfin versions the client can report to the server which Dolby Vision profiles it supported.

Pentaphon:
[...]

Issue #551, opened 2025-12-23
[FEA] - Setting to disable Dolby Vision support


en6ads:
### Description

Not all exoplayer based clients can handle Dolby Vision P7 MEL and FEL correctly.

For example see this issue with the Fire Stick 4k:
https://forum.jellyfin.org/t-solved-black-screen-when-playing-4k-hdr-content-on-fire-stick-4k?pid=40241#pid40241
https://github.com/damontecres/Wholphin/issues/545

Please add a playback setting to disable / enable Dolby Vision Support, so this tells the server it does / does not support Dolby Vision, and therefore server can remux Dolby Vision files to HDR when needed.

Many thanks.

damontecres:
Related to #245 & #244

PR #1876, opened 2026-08-20
Add option to prefer Dolby Vision over HDR10+ on hybrid videos

Jycreyn:
## Description

Adds a "Prefer Dolby Vision over HDR10+" switch (off by default) under Experimental settings. That screen only shows up once "Experimental settings" is enabled under Advanced, and until then the switch is read as off: it goes through `experimentalPreferences.get {}`, so nothing is wrapped and the playback path is untouched.

Most current 4K web releases carry both Dolby Vision (profile 8.1) RPUs and HDR10+ dynamic metadata in the same HEVC stream. Some MediaTek based devices (Philips and TCL TVs, some Fire TV models) latch onto the HDR10+ metadata on those files and never engage Dolby Vision, even though a Dolby Vision decoder is selected.

I isolated this on my Philips 55OLED810 by playing the same episode three ways, direct play in all three cases:

| | Stream | TV shows |
|---|---|---|
| A | MKV with DV 8.1 + HDR10+ | HDR10+ |
| B | Same MKV with the HDR10+ metadata removed using [hdr10plus_tool](https://github.com/quietvoid/hdr10plus_tool), DV untouched | Dolby Vision |
| C | Same stream remuxed to MP4, both kinds of metadata present | HDR10+ |

The container doesn't matter and the Dolby Vision signaling is picked up correctly. The only thing the app can do is keep the HDR10+ metadata away from the decoder.

When the switch is on, and only when a Dolby Vision decoder was actually selected, the HDR10+ SEI messages (SMPTE ST 2094-40, recognized by their ITU-T T.35 header `B5 003C 0001`) are masked in each access unit before it is queued to MediaCodec. The SEI payload type byte is overwritten with an unused value (254), so the decoder skips the message. That is one byte rewritten in place per message: the masking filter returns the size it was given, so the access unit is queued exactly as long as it came out of the extractor, which is what makes it safe with both the synchronous and asynchronous MediaCodec adapters. media3 does the same thing for HAGC metadata in `HevcSeiUtil`; this filter only targets ST 2094-40, so other T.35 SEI such as CEA-708 captions are not affected. The filter was written from the H.265 spec, no Kodi or media3 code was copied.

The hook is generic rather than HDR10+ specific, as asked in the thread. `BitstreamFilteringCodecAdapterFactory` wraps the codec adapter with a `List<BitstreamFilter>`, and a filter is `(ByteBuffer, offset, size) -> Int` where the returned `Int` is the new size of the access unit: the adapter queues the size the chain reports, and hands each filter the size the previous one reported. A filter which only overwrites bytes returns the size unchanged, which is what the HDR10+ masking does; a filter which drops NAL units compacts what it keeps and reports the shorter size. A reported size which is negative or larger than the one the filter was given is ignored rather than queued. That is what lets the Dolby Vision profile 7 to 8.1 conversion of #244 arrive later as one more filter instead of a second copy of the adapter plumbing.

When the switch is off, the playback path is unchanged. On a device without a Dolby Vision decoder (or a display without DV support) the wrapper is never installed, so HDR10+ only devices keep their metadata. Encrypted samples, codec config buffers and empty buffers are passed through untouched.

Kodi has an equivalent option for the same reason: xbmc/xbmc@032cbfeb and xbmc/xbmc@e2c54e7f.

### Related issues

Implements what was requested in #1855. The filter list is what #244 (Dolby Vision profile 7 direct play) needs, so that conversion can be added as another filter. #1895 implemented the same masking independently and was closed in favour of this PR; the device profile part of it was not carried over, see the note under Testing. Upstream: androidx/media#1895, where the maintainers say the bug is in the device codecs and that they don't plan a workaround in ExoPlayer itself.

### Testing

Unit tests, 27 in total:

- 14 for the SEI filter with synthetic NAL units: multi-message SEI NALs, emulation prevention bytes, ff-extended payload types, 3 and 4 byte start codes, CEA-708 T.35 messages, truncated NALs, and length-prefixed buffers (left untouched; media3's extractors always emit Annex B).
- 8 for the filter chain: the size a filter reports is what comes out, each filter is handed the size the previous one reported, a filter which drops bytes and compacts a direct buffer, and negative or oversized reports are ignored.
- 5 for the hand off to the decoder: the size the chain reports is the size which gets queued, and codec config buffers, empty buffers and a missing input buffer are forwarded untouched.

Tested independently by @SergeiAK on hardware I don't have, on the `2820c724` build, against a 10.11.11 server: Homatics Box R 4K Plus (Amlogic S905X4) + LG OLED55BX, and Nvidia Shield Pro 2019 + TCL 65C825, both outside `KnownDefects.hevcDoviHdr10PlusBug`. Switch off, neither log line appears and playback is unchanged. Switch on with Dolby Vision but no HDR10+, the wrapper is installed, reports 0 masked messages and the display stays in Dolby Vision. On a hybrid title it masks about one message per frame (1103 and 964 over 40 seconds). Dual layer profile 7 files pass through the filter unharmed, 0 masked when they carry no HDR10+. The wrapper attached on `c2.amlogic.dolby-vision.dvhe.decoder` (Codec2) and on `OMX.Nvidia.DOVI.decode` (OMX), so the mime gating holds on two different vendor stacks. Full numbers in [his comment](https://github.com/damontecres/Wholphin/pull/1876#issuecomment-5573012201).

On my Philips 55OLED810 (MediaTek) with Jellyfin 10.11: a DV8.1+HDR10+ hybrid (checked with ffprobe: DV RPU plus per-frame ST 2094-40) direct plays with the Dolby Vision banner when the switch is on. With the switch off it plays as HDR10+ like before. A DV-only file behaves the same as before, and seeking and jumping to the next episode work normally.

**Picking test material.** On a 10.11 server the server tells the range types apart itself, so this is a query rather than an ffprobe run:

```
/Items?includeItemTypes=Movie&recursive=true&fields=MediaStreams
```

then filter the video stream on `VideoRangeType`, the hybrid ones being `DOVIWithHDR10Plus` and `DOVIWithELHDR10Plus`, with `DvProfile`, `DvBlSignalCompatibilityId` and `ElPresentFlag` alongside to confirm what each file actually is. On a 10.10 server that distinction is not reported and the file has to be checked directly, the RPU being stream level side data and the HDR10+ messages per frame:

```
ffprobe -v error -select_streams v:0 -show_streams <file> | grep -i dovi
ffprobe -v error -select_streams v:0 -read_intervals '%+#120' -show_frames <file> | grep -i hdr_plus
```

**Getting it to the decoder.** The file also has to reach the decoder untranscoded, which is where one caveat matters: on the Fire TV models in `KnownDefects.hevcDoviHdr10PlusBug` (AFTKRT, AFTKA, AFTKM) against a Jellyfin 10.11 server, `DeviceProfileUtils` marks `DOVIWithHDR10Plus` as an unsupported range type, so the server transcodes the hybrid file and the access units never reach MediaCodec as DV+HDR10+. On those sticks this switch cannot change anything, whatever the file. That gating is #1895's other half and is deliberately not touched here, since it sits on the same line as #1889 and is device profile policy rather than a bitstream hook. A valid test needs either a device outside that list (any MediaTek TV, which is the case in #1855), a 10.10 server, or a local build with that marker removed.

Two log lines say what happened, both from the codec adapter: `Filtering the bitstream of Dolby Vision playback on <decoder>` when the wrapper is installed, and on release `Releasing bitstream filtering adapter, HDR10+ SEI masking (N messages masked)` with the count of what it actually masked. If the first line is missing, either the switch is off, experimental settings are off, or the selected decoder is not a Dolby Vision one.

## Screenshots

Prefer Dolby Vision over HDR10+ on the experimental settings screen

## AI or LLM usage

The implementation (bitstream filter, MediaCodecAdapter wiring, unit tests) were written with AI assistance (Claude). I reviewed the changes and validated the behavior myself on a real MediaTek TV (Philips 55OLED810), with the setting on and off, on hybrid and DV-only content.


SergeiAK:
Different chip, different metadata, same remedy — and that is why I would like to argue for
keeping this wrapper generic rather than HDR10+-specific.

I ran into the mirror image of your case on Amlogic. Homatics Box R 4K Plus (S905X4,
Android 14) + LG OLED55BX: dual-layer Profile 7 remuxes whose RPU had been converted to 8.1
in band still played as HDR10 — `onHdrInfoChange=1`. The reason is not the RPU but the
enhancement layer left in the access unit: the NAL type 63 units are still there, so the
bitstream is still dual layer and the device keeps its Dolby Vision pipeline out of it, even
though `Format.codecs` now reads `dvhe.08…`. Blank those type 63 units before the sample
reaches the decoder and the same file on the same build gives `onHdrInfoChange=3` and the
Dolby Vision banner on the TV's own info panel. Control: a native single-layer 8.1 file opens
the same decoder and gives `=3` either way, so the decoder selection is not the variable.

Census of a 75-frame `ffmpeg -c copy` extract, i.e. the exact bytes media3 sees — 265 type 63
units against 70 slices and 75 RPUs, ordered `… 63 63 63 62 | 35 39 1 …` per access unit:

```
type  1: 70    slices
type 20:  5    IDR
type 35: 75    AUD
type 39: 110   SEI
type 62: 75    RPU
type 63: 265   enhancement layer
```

This is running in production, not a proposal: [Just+ Player](https://github.com/just-plus-player/just-plus-player)
v1.2.0 does the type 63 removal before rewriting the RPU, and Profile 7 files reach the display
as Dolby Vision on both of my boxes. Wholphin's own request for it is #244.

The two cases are the same shape. Yours masks an ST 2094-40 SEI by overwriting its payload
type; the Profile 7 case turns type 63 NAL units into filler. Both rewrite bytes **in place**
inside the access unit, both are size-preserving for exactly the reason you give in the PR
description, and both want the hook you just built — a `MediaCodecAdapter` wrapper that sees
Annex B after the extractor and before MediaCodec.

So, a request rather than a criticism: if this lands, could the wrapper hold a list of filters
instead of one? Concretely, `Hdr10PlusMaskingCodecAdapterFactory` becomes something like
`BitstreamFilteringCodecAdapterFactory(delegate, filters)` where a filter is
`(ByteBuffer, offset, size) -> Int`, and `Hdr10PlusSeiFilter` is the first entry. Identical
behaviour with a single filter installed, identical gating on
`codecInfo.mimeType == VIDEO_DOLBY_VISION`, no extra cost — but the Profile 7 conversion then
arrives as one more filter plus a preference, instead of a second copy of the adapter plumbing
next to yours. If you would rather keep this PR minimal and land it as is, that is entirely
reasonable too; I can propose the extraction separately afterwards.

Happy to test this on hardware you do not have: Amlogic S905X4 (Homatics) and Nvidia Shield Pro
2019, neither of them MediaTek. That will not reproduce your symptom, but it does cover the two
paths that matter for a merge — switch off leaves playback unchanged, and DV-only files behave
the same with the switch on.

Jycreyn:
> Different chip, different metadata, same remedy — and that is why I would like to argue for keeping this wrapper generic rather than HDR10+-specific.
> 
> I ran into the mirror image of your case on Amlogic. Homatics Box R 4K Plus (S905X4, Android 14) + LG OLED55BX: dual-layer Profile 7 remuxes whose RPU had been converted to 8.1 in band still played as HDR10 — `onHdrInfoChange=1`. The reason is not the RPU but the enhancement layer left in the access unit: the NAL type 63 units are still there, so the bitstream is still dual layer and the device keeps its Dolby Vision pipeline out of it, even though `Format.codecs` now reads `dvhe.08…`. Blank those type 63 units before the sample reaches the decoder and the same file on the same build gives `onHdrInfoChange=3` and the Dolby Vision banner on the TV's own info panel. Control: a native single-layer 8.1 file opens the same decoder and gives `=3` either way, so the decoder selection is not the variable.
> 
> Census of a 75-frame `ffmpeg -c copy` extract, i.e. the exact bytes media3 sees — 265 type 63 units against 70 slices and 75 RPUs, ordered `… 63 63 63 62 | 35 39 1 …` per access unit:
> 
> ```
> type  1: 70    slices
> type 20:  5    IDR
> type 35: 75    AUD
> type 39: 110   SEI
> type 62: 75    RPU
> type 63: 265   enhancement layer
> ```
> 
> This is running in production, not a proposal: [Just+ Player](https://github.com/just-plus-player/just-plus-player) v1.2.0 does the type 63 removal before rewriting the RPU, and Profile 7 files reach the display as Dolby Vision on both of my boxes. Wholphin's own request for it is #244.
> 
> The two cases are the same shape. Yours masks an ST 2094-40 SEI by overwriting its payload type; the Profile 7 case turns type 63 NAL units into filler. Both rewrite bytes **in place** inside the access unit, both are size-preserving for exactly the reason you give in the PR description, and both want the hook you just built — a `MediaCodecAdapter` wrapper that sees Annex B after the extractor and before MediaCodec.
> 
> So, a request rather than a criticism: if this lands, could the wrapper hold a list of filters instead of one? Concretely, `Hdr10PlusMaskingCodecAdapterFactory` becomes something like `BitstreamFilteringCodecAdapterFactory(delegate, filters)` where a filter is `(ByteBuffer, offset, size) -> Int`, and `Hdr10PlusSeiFilter` is the first entry. Identical behaviour with a single filter installed, identical gating on `codecInfo.mimeType == VIDEO_DOLBY_VISION`, no extra cost — but the Profile 7 conversion then arrives as one more filter plus a preference, instead of a second copy of the adapter plumbing next to yours. If you would rather keep this PR minimal and land it as is, that is entirely reasonable too; I can propose the extraction separately afterwards.
> 
> Happy to test this on hardware you do not have: Amlogic S905X4 (Homatics) and Nvidia Shield Pro 2019, neither of them MediaTek. That will not reproduce your symptom, but it does cover the two paths that matter for a merge — switch off leaves playback unchanged, and DV-only files behave the same with the switch on.

Thanks, that's a useful data point. The type 63 units staying in the access unit
explains why fixing up the codecs string alone doesn't get you there, and it matches
what I ended up with on my side: the decoder selection was never the variable either.

No objection to the filter list. maskHdr10PlusSeiMessages already has the
(ByteBuffer, offset, size) -> Int shape you describe, so it's mostly a rename and a
loop, with no behaviour change when a single filter is installed. I'd rather not grow
this PR before @damontecres has had a look at it though. If he'd prefer the generic
version I'll do it here, otherwise your separate extraction afterwards sounds like the
cleaner path.

I'll take you up on the testing. The two paths that matter for a merge are the switch
off leaving playback untouched, and DV-only files behaving the same with it on, an
S905X4 and a Shield cover both. The branch is feature/prefer-dv-over-hdr10plus on my
fork if you want to build it; the setting is under ExoPlayer options and is off by
default.

damontecres:
One change needed would be to move this to experimental settings.

That said, I haven't fully reviewed this yet and honestly, it's probably a bit over my head at this low of a level. Are there are resources you can recommend so I can learn more?

I see there are a number of hardcoded byte and other constants, so understanding where they come from would be helpful.

SergeiAK:
Not my PR, so this isn't a review note, just an attempt to help with the question about the
constants. I went looking for the same thing while reading the diff — here is where it led me.
Correct me where I got it wrong.

As far as I can tell none of these numbers are invented: each one is a field value from a
spec, and the documents you need are free.

Most of it is in **ITU-T H.265** (https://www.itu.int/rec/T-REC-H.265):

| In the code | Where to look |
|---|---|
| `(byte shr 1) and 0x3F` | §7.3.1.2 — `nal_unit_type` is 6 bits, right after `forbidden_zero_bit` |
| `39` prefix SEI, `<= 31` VCL | Table 7-1, the list of NAL unit types |
| skipping `00 00 03` | §7.4.2 — emulation prevention |
| scanning for `00 00 01` | Annex B — byte stream format |
| `while (b == 0xFF) payloadType += 255` | §D.2.1 — `sei_message` syntax |
| payload type `4` | §D.2.1 — `user_data_registered_itu_t_t35` |
| why `254` gets through | §D.3.1 — a decoder must ignore SEI payload types it doesn't know |

The `0xB5` byte is from **ITU-T T.35** (https://www.itu.int/rec/T-REC-T.35), where it is the
country code for the United States. The `0x003C` / `0x0001` pair is quickest to check against
FFmpeg: [`libavcodec/itut35.h`](https://github.com/FFmpeg/FFmpeg/blob/master/libavcodec/itut35.h)
declares `ITU_T_T35_COUNTRY_CODE_US 0xB5` and `ITU_T_T35_PROVIDER_CODE_SAMSUNG 0x003C` outright.
The spec for the payload itself (SMPTE ST 2094-40) is paid, and I haven't bought it, but I don't
think it's needed to review this.

Probably the most useful reading is closest to home: media3 parses the same bitstream in
`NalUnitUtil` (`androidx.media3.container`) and `H265Reader` (`androidx.media3.extractor.ts`).
Same start code scan, same emulation prevention, same NAL header — except it's code the app
already ships.

Types 62 (RPU) and 63 (enhancement layer), which I mentioned above, aren't from H.265 but from
Dolby: "Dolby Vision Profiles and Levels" and "Dolby Vision Streams Within the ISO Base Media
File Format". Both are published free on Dolby's professional site and are easy to find by
title — no direct link from me, since their CDN doesn't load here and I'd rather not hand you
a stale URL.

One small thought, mostly for @Jycreyn: putting the clause number in the KDoc of each constant,
and using media3's own constant where one exists, would let the next reader get this answer
straight from the diff.

Jycreyn:
Done, it's under experimental settings now. Reads through
experimentalPreferences.get {} so it does nothing while experimental is off.

On the constants: SergeiAK's list is right, none of them are invented. It's all
H.265, which the ITU gives away for free. nal_unit_type is 7.3.1.2, the type
numbers are Table 7-1, start codes are Annex B, the 00 00 03 skip is 7.4.2, and
the SEI parse loop plus payload type 4 are both D.2.1.

The clause that actually matters for reviewing this is D.3.1: a decoder has to
ignore payload types it doesn't know. That's the whole trick. 254 isn't magic,
it's just a value nothing is assigned to, so the decoder skips the message
instead of acting on it. Size doesn't change, nothing gets removed.

0xB5 is the T.35 country code for the US, 0x003C is Samsung (HDR10+ is theirs).
Both are in FFmpeg's libavcodec/itut35.h if you'd rather check code than spec.
ST 2094-40 itself is paywalled, but you don't need it here, the filter never
reads past those 5 bytes.

If you'd rather read code than spec generally, media3's own NalUnitUtil and
H265Reader do the same scan and they already ship in the app. Probably the
fastest way in.

Anyway, I've put the clause in the KDoc of each constant so you can get this
from the diff instead of from me, and the prefix SEI type is
NalUnitUtil.H265_NAL_UNIT_TYPE_PREFIX_SEI now. The T.35 ones have no media3
equivalent so they stayed local with the FFmpeg reference in the comment.

One more thing I should have mentioned, since it's the shortest answer to your
question: media3 already does this. androidx.media3.exoplayer.video.HevcSeiUtil
has SEI_PAYLOAD_TYPE_UNSPECIFIED = 254 and uses it the same way, and its public
stripAllT35Metadata does the blunt version, every T.35 message in the buffer.

That's the reason this filter exists rather than a call to that: stripping all
T.35 would take CEA-708 captions with it, since they ride the same mechanism.
This one only matches the ST 2094-40 provider code and leaves everything else
alone.

SergeiAK:
FYI: #1895 opened today implements the same payload-type-4-to-254 masking independently,
with the filter hooked into the video renderer rather than the codec adapter, and it also
adjusts the device profile so the server stops transcoding. Neither PR referenced the other,
so linking them here — worth settling on one hook before either lands, especially with the
profile 7 → 8.1 case from #244 waiting for the same machinery.

damontecres:
> No objection to the filter list. maskHdr10PlusSeiMessages already has the (ByteBuffer, offset, size) -> Int shape you describe

I think making this change in this PR is a good idea.

I believe one of the Fire TV sticks I have is affected by this issue, so I just need to find an affected video in my library to test these changes. Also very appreciative if others can test as well.

damontecres:
> worth settling on one hook before either lands, especially with the profile 7 → 8.1 case from https://github.com/damontecres/Wholphin/issues/244 waiting for the same machinery

I believe the approach in this PR will be compatible with the conversion library (#1540) since that changes the `MediaSourceFactory`. And I prefer this PR over #1895

SergeiAK:
Glad the filter list is going in — thanks for taking it.

One small request about the filter signature while it is still being written.

The HDR10+ masking never changes the sample size: one payload type byte is
rewritten in place, and the buffer stays exactly as long as it was. The DV
Profile 7 → 8.1 conversion from #244 is not like that — the enhancement layer
is dropped, so the sample comes out shorter than it went in.

So could the filter be allowed to report a new length, and could the caller
actually use what it returns? Something like

    (ByteBuffer, offset, size) -> Int

where the returned Int is the new size. For `Hdr10PlusSeiFilter` it is always
the size it was given, so nothing changes for this PR. But if the contract ends
up being "edit in place, length stays the same", then the conversion cannot be
added later as just another filter, and it would need its own copy of the same
plumbing — which is the thing we were trying to avoid.

I am testing that conversion on Amlogic right now (with the library author, in
suyash192/ExoplayerHdrUtils#2), so I will know soon enough whether it can ride
on this hook. Happy to be wrong about the details — I just wanted to flag it
before the interface is settled.

SergeiAK:
One practical question: do you have a debug APK of this branch you could share?

I offered to test on Amlogic S905X4 and Shield Pro, and I would rather not guess at the build — the PR workflow builds the APK but does not upload it as an artifact, so there is nothing to download from CI, and there is no release on your fork.

If it is easier for you not to, no problem at all, I will build it myself. Test files are not an issue — I have Profile 7 and Profile 8 material in my own library.

Jycreyn:
Here you go: https://github.com/Jycreyn/Wholphin/releases/tag/pr1876-ae98b3ad

arm64-v8a and a universal APK, both built from ae98b3ad, so it's the PR head as it is now, not a rebase.

It's a debug build signed with my own debug key, so it installs as `com.github.damontecres.wholphin.debug` and sits next to a normal install instead of updating it.

The switch is under experimental settings, so you have to turn those on first, otherwise it reads as off and the adapter never gets wrapped. Off by default once you're there.

One thing about timing: the branch is going to change for the filter list, so this build is the "before" state. I'll put another one in the same place when that's done, in case you'd rather test once.

Jycreyn:
Pushed the filter list (2820c724).

`Hdr10PlusMaskingCodecAdapterFactory` is now `BitstreamFilteringCodecAdapterFactory(delegate, filters)` and the HDR10+ masking is the single entry we install. Same gating on `codecInfo.mimeType == VIDEO_DOLBY_VISION`, same behaviour with one filter.

@SergeiAK on the size: the `Int` a filter returns is the new size of the access unit, and `queueInputBuffer` forwards that size to the decoder instead of the one it was called with. So a filter that drops NAL units and compacts what it keeps is fine, no second copy of the plumbing. Each filter is handed the size the previous one reported. A size that is negative or larger than the one the filter was given is ignored rather than queued, since that would hand the decoder a truncated or over long access unit. The masking returns the size unchanged, so nothing moves for this PR. The contract is written on `BitstreamFilter`, tell me if the wording doesn't cover what the profile 7 conversion needs.

Tests: 8 on the chain (size forwarded, filters chained, bad values ignored) on top of the 13 on the masking itself.

New APK in the same release, both builds are there now: https://github.com/Jycreyn/Wholphin/releases/tag/pr1876-ae98b3ad

@damontecres that is the change you asked for on the 5th.

Jycreyn:
Description rewritten and the screenshot retaken. Both still showed the switch in the ExoPlayer options, which is where it was before ae98b3ad, so the PR page was contradicting the one change you asked for. The description now also covers the filter list, the size contract, and how to tell from the logs whether the wrapper was installed.

@damontecres one thing to know before you test on the Fire TV stick, because the switch will look like it does nothing there. On AFTKRT, AFTKA and AFTKM (`KnownDefects.hevcDoviHdr10PlusBug`) against a Jellyfin 10.11 server, `DeviceProfileUtils` puts `DOVIWithHDR10Plus` in the HEVC unsupported range types, so the server transcodes the hybrid file and the access units never reach MediaCodec as DV+HDR10+. No bitstream filter can do anything about that. A valid test needs a device outside that list (any MediaTek TV, like the reporter in #1855), a 10.10 server, or a local build with that marker removed.

That gating is the other half of #1895 and I left it out on purpose: it is the same line #1889 is already editing, and it is device profile policy rather than a bitstream hook. If you confirm on hardware that masking the SEI is enough on those sticks, I will open the one line gate plus the `DeviceProfileUtilsTest` from #1895 as its own PR.

Also, the CI run on the head commit is sitting in `action_required`, so the PR shows no checks at all. It needs your approval to run.

@SergeiAK the hand off you asked about has a regression guard now. `BitstreamFilteringCodecAdapter` is internal and five tests pin that the size the chain reports is the size `queueInputBuffer` forwards, and that codec config buffers, empty buffers and a missing input buffer go through untouched. The chain test also drops bytes for real over a direct buffer, since that is what MediaCodec hands a filter: `hasArray()` is false, `array()` throws, so absolute get and put only, and the absolute bulk `put(int, ByteBuffer, int, int)` is API 34 while `minSdk` is 23. That is on the `BitstreamFilter` contract now rather than something the profile 7 filter has to find out on device. 27 tests in total.

The APK in the release is still the 2820c724 one: fce8eb9a is tests, a KDoc and a default value in `AppPreferencesSerializer`, no behaviour change.

SergeiAK:
Tested on two devices, both outside `KnownDefects.hevcDoviHdr10PlusBug`, against a
Jellyfin 10.11.11 server. Build: the `2820c724` APK from your release, so the filter
list version, not the one before it.

- Homatics Box R 4K Plus (Amlogic S905X4, Android 14, 32 bit userspace) + LG OLED55BX
- Nvidia Shield Pro 2019 (Android 11) + TCL 65C825

All files direct play, all runs are 40 seconds from the start of the title.

### Amlogic S905X4 + LG BX

| File | Switch | Wrapper installed | Masked | Display |
|---|---|---|---|---|
| DV 8.1, no HDR10+ | off | no | — | `DolbyVision-Std` |
| DV 8.1, no HDR10+ | on | yes | 0 | `DolbyVision-Std` |
| DV 8.1 + HDR10+ | off | no | — | `DolbyVision-Std` |
| DV 8.1 + HDR10+ | on | yes | 1103 | `DolbyVision-Std` |
| DV 7 FEL | on | yes | 0 | HDR10 |
| DV 7 FEL + HDR10+ | on | yes | 946 | HDR10 |

Decoder `c2.amlogic.dolby-vision.dvhe.decoder`, `onHdrInfoChange` 1 → 3, display state
read from `/sys/class/amhdmitx/amhdmitx0/hdmi_hdr_status`.

### Shield Pro 2019 + TCL C825

| File | Switch | Wrapper installed | Masked |
|---|---|---|---|
| DV 8.1, no HDR10+ | off | no | — |
| DV 8.1, no HDR10+ | on | yes | 0 |
| DV 8.1 + HDR10+ | on | yes | 964 |
| DV 7 FEL | on | yes | 0 |
| DV 7 FEL + HDR10+ | on | yes | 946 |

Decoder `OMX.Nvidia.DOVI.decode`, with `ACodec: Setting Dolby Vision Specific attributes
to Video Decoder` and `DisplayManagementConfig: DolbyVision Source in VPM Mode` on every
run.

### What this covers

**Switch off: nothing changes.** Neither log line appears at all on either device, and
playback is what it was before the build.

**Switch on, Dolby Vision without HDR10+: nothing breaks.** The wrapper is installed on
every stream where a Dolby Vision decoder was selected, so ordinary DV goes through the
filter too. It reports 0 masked messages and the display stays in Dolby Vision. That is
the case I was most concerned about, since the symptom itself does not reproduce on this
hardware.

**The masking works on real material, not only on synthetic NAL units.** 1103 and 964
messages over 40 seconds of a hybrid title is about one per frame, which is what per
frame ST 2094-40 should look like.

**The hook attaches by mime, on two different codec stacks.** Codec2 on Amlogic and OMX
on Nvidia, different vendors and different names, and `codecInfo.mimeType ==
VIDEO_DOLBY_VISION` picked up both. Worth stating explicitly since the gating is the one
thing every future filter in the list will inherit.

**Profile 7 passes through unharmed, and this is the part I care about most.** With
`Direct play Dolby Vision Profile 7` enabled the two layer files reach the decoder with
their enhancement layer in band, so the filter walks access units containing NAL types 62
and 63. Nothing is corrupted: the clean P7 file reports 0 masked, the P7 file that also
carries HDR10+ reports 946, and both play normally on both boxes.

What the same runs also show is why I am in this part of the tracker at all. The two
devices disagree about profile 7, with the same app, the same build and the same file:

| | Profile 7 FEL | Profile 8.1 |
|---|---|---|
| Shield Pro 2019 + TCL C825 | Dolby Vision on the TV | Dolby Vision |
| Amlogic S905X4 + LG BX | **HDR10** — `onHdrInfoChange = 1`, `hdmi_hdr_status = HDR10-GAMMA_ST2084` | Dolby Vision — `1 → 3`, `DolbyVision-Std` |

Shield handles profile 7 itself, so nothing is needed there. On the Amlogic box the file
direct plays, a Dolby Vision decoder is selected, and the display still ends up in HDR10,
because the enhancement layer and the RPU are never converted. That is #244, where I have just posted these numbers in full
(https://github.com/damontecres/Wholphin/issues/244#issuecomment-5573010408), and it is
every profile 7 Blu-ray remux in my library. The same box plays the same file as Dolby
Vision in [Just+ Player](https://github.com/just-plus-player/just-plus-player) since
v1.2.0 ([PR #120](https://github.com/just-plus-player/just-plus-player/pull/120):
conversion through an MIT licensed library from Maven Central plus one converter class
that handles the in band enhancement layer), so it is not an Amlogic limitation either.

Which is why your filter list matters beyond HDR10+. The chain takes
`(ByteBuffer, offset, size) -> Int` and applies the size a filter reports, and that is
exactly what a profile 7 to 8.1 conversion needs, because dropping the enhancement layer
makes the access unit shorter. The measurements above say the chain already survives real
two layer material, before anyone writes that filter. I would rather see the conversion
land as one more entry in your list than as a second copy of the same plumbing elsewhere,
and I am happy to write it and test it on this hardware.

### One correction to the testing notes in the description

> Jellyfin reports DV-only and DV+HDR10+ alike as `DOVIWithHDR10`, so the info panel
> says "Dolby Vision" either way

On a 10.11 server this is no longer true: the server distinguishes them itself and
returns `VideoRangeType` values of `DOVIWithHDR10`, `DOVIWithHDR10Plus`, `DOVIWithEL` and
`DOVIWithELHDR10Plus`. Picking test material is a query rather than an ffprobe run:

```
/Items?includeItemTypes=Movie&recursive=true&fields=MediaStreams
```

then filter on the video stream's `VideoRangeType`, with `DvProfile`,
`DvBlSignalCompatibilityId` and `ElPresentFlag` alongside it to confirm what each file
actually is. That is how the six files above were chosen. Someone testing on a 10.10
server still needs ffprobe.

### Unrelated to this PR, noted for the record

On the Amlogic box, both profile 7 files select `c2.amlogic.dolby-vision.dav1.decoder`,
the AV1 branch of the vendor's Dolby Vision decoders, for an HEVC stream. The P8 files on
the same box select `c2.amlogic.dolby-vision.dvhe.decoder`. Shield has a single DV
decoder and shows nothing of the sort, so this looks like decoder selection on a device
that exposes several. Playback works, so it may be harmless. Mentioning it only because
it turned up while testing this branch; it is not caused by it.

### One packaging note

The arm64 APK does not install on Amlogic boxes with a 32 bit userspace:

```
INSTALL_FAILED_NO_MATCHING_ABIS: Failed to extract native libraries, res=-113
ro.product.cpu.abilist = armeabi-v7a,armeabi
```

The Homatics box above is one of those, and a good share of the S905X4 and S905Y4
devices Wholphin runs on are too. The universal APK installs and is what all the Amlogic
numbers here come from. Worth keeping the universal build in the release for any future
test drop rather than arm64 alone, otherwise the devices most affected by device profile
and Dolby Vision issues are the ones which cannot install it.

Jycreyn:
That is a lot more than I could test myself, thank you.

You are right about the range types and I have fixed the description: on 10.11 the server distinguishes them, so picking material is the `/Items` query with `VideoRangeType` (plus `DvProfile`, `DvBlSignalCompatibilityId` and `ElPresentFlag`), and ffprobe is only needed on 10.10. Your numbers are now summarised in the Testing section with a link to your comment, since that is what a reviewer reads at merge time.

Two results matter more than the rest, to me. The wrapper attaching on `c2.amlogic.dolby-vision.dvhe.decoder` and on `OMX.Nvidia.DOVI.decode` means the mime gating holds on two vendor stacks, which is the one thing every future filter in the list inherits. And profile 7 dual layer files walking through the filter with 0 masked and playing normally is the case I had no way to exercise: the filter now provably does not trip over NAL types 62 and 63 in band.

On the conversion: yes, please write it as another entry in the list. Since your last message I pushed fce8eb9a, which adds the piece that was missing for that: five tests pin that the size the chain reports is the size `queueInputBuffer` forwards to the decoder, so if a filter shortens an access unit and the adapter ever stops honouring it, a test fails rather than a device. The contract on `BitstreamFilter` also spells out what a dropping filter has to do now, including that the buffer is the decoder's direct input buffer, so `array()` throws and compaction means absolute puts or a `duplicate()`, with the absolute bulk `put(int, ByteBuffer, int, int)` being API 34 while `minSdk` is 23.

On packaging, agreed, and it is not only Amlogic: my own Philips 55OLED810 is `armeabi-v7a` as well and refused the arm64 split with the same `res=-113`. The universal APK stays in the release for every test drop.

The `dav1` decoder picked for an HEVC profile 7 stream is worth its own issue if it turns out to matter, but I agree it is not from this branch.

xdlolrgb:
Would this be able to help the player fall back to HDR10 if the video playing is DV HDR10 and the device doesn't support DV?

Jycreyn:
@xdlolrgb Not on its own, no. The filter only runs once a Dolby Vision decoder has already been
selected: the wrapper is installed only when `configuration.codecInfo.mimeType ==
VIDEO_DOLBY_VISION`, and all it then does is rewrite one byte in place, the payload type
of an HDR10+ SEI message. On a device with no DV decoder nothing is wrapped and the filter
never runs. It also does not touch decoder selection, track selection, the `Format`, or the
device profile sent to the server, so it cannot change what a non-DV device does with a DV
file.

The good news is that the fallback you are describing already exists, in two places, and
neither of them is this PR.

**The device profile.** With no DV decoder on the device, `DeviceProfileUtils` declares
profile 5 (`DOVI`) unsupported unconditionally, but declares profile 8.1
(`DOVIWithHDR10`) unsupported *only* if the device also lacks HEVC HDR10. So on an
HDR10-capable device without Dolby Vision, a profile 8.1 file is advertised as supported
and direct plays.

**Media3's own decoder fallback.** In media3 1.11.0
`MediaCodecUtil.getAlternativeCodecMimeType` maps DV profile 8 (and 4) to `video/hevc`,
9 to `video/avc` and 10 to `video/av01`; profiles 5 and 7 map to nothing. And on API 26+,
when the display does not advertise Dolby Vision,
`MediaCodecVideoRenderer.getDecoderInfos` returns *only* those alternatives. A profile 8.1
stream therefore lands on a plain HEVC Main10 decoder, which decodes the HDR10-compatible
base layer and ignores the RPU — which is exactly the fallback you are after.

So it depends on the file:

- **Profile 8.1** (`VideoRangeType` = `DOVIWithHDR10`, which is usually what "DV HDR10"
  means): it should already play as HDR10 on a device with no DV. If it does not on yours,
  that is device profile or decoder selection rather than the bitstream, and
  #551 is the thread for it — worth saying which device and
  what you get instead.
- **Profile 5**: there is no HDR10 base layer to fall back to, the RPU is not optional, so
  the server transcodes. Nothing client side changes that.
- **Profile 7**: no base-layer mime either, and the enhancement layer would have to be
  converted first. That is #244.

Happy to dig in if it is the first case, since that is the one where something is actually
going wrong.

alexkahler:
Good explanation @Jycreyn

damontecres:
Now that I got `v1.0.8` out, hopefully I can test this PR and get it merged this week!

fanni42:
I'm waiting too, because after this if you could improve the playback of hybrid Profile 7 to play as Dolby Vision on the Amlogic S905X4, that would be great.

SergeiAK:
@fanni42 That part is already written and measured — it sits on top of this branch as an entry in the `BitstreamFilter` list this PR adds.

On an Amlogic S905X4 a Profile 7 FEL remux direct plays in Dolby Vision: `dvhe.07.06` → `dvhe.08.06`, `onHdrInfoChange` 1 → 3, 1200 access units converted, 0 refused. Numbers and the two things that turned out to matter more than the RPU rewrite itself are in https://github.com/damontecres/Wholphin/issues/244#issuecomment-5694869621.

I'm holding the PR until this one lands, since the filter has nowhere to plug in before then.

Jycreyn:
Just a usage data point, not a nudge. I've been running this branch as the only Wholphin build on my TVs since I opened it in August, including a Philips 55OLED810 (Google TV). Daily household use, rebased on main every time, and it hasn't caused a single regression here. Hybrid DV/HDR10+ titles play in Dolby Vision, and everything else behaves exactly like stock, which is what you'd expect: the wrapper is only installed when the selected decoder mime type is `VIDEO_DOLBY_VISION`, so on non-hybrid content the filter never runs at all.

A signed test build of this branch's head has also been in the hands of someone else from this thread since early September, so it isn't only my hardware.

CI is green and the branch still merges cleanly. Happy to rebase or answer anything whenever it suits you.

PR #1895, opened 2026-08-23
Add Fire TV hybrid Dolby Vision workaround

hinneslung:


SergeiAK:
Not a review, just a heads-up: this PR and #1876 were opened three days apart and neither
references the other. Both neutralize the HDR10+ SEI message the same way — payload type 4
rewritten to the reserved 254 — but hook it in at different points: #1876 wraps the
`MediaCodecAdapter`, this one filters in the video renderer. #1889 and #1644 touch the same
area from other angles.

Four open PRs around Dolby Vision, written independently, is a good moment to compare them
and settle on one hook rather than merge whichever gets reviewed first.

One thing to weigh when choosing: the same machinery is wanted for converting Dolby Vision
profile 7 to 8.1, asked for in #244 since November 2025 — and that one is already working
elsewhere, [Just+ Player](https://github.com/just-plus-player/just-plus-player) shipped it in
v1.2.0. Whichever hook wins here will decide whether that arrives as one more filter or as a
third copy of the same code.

damontecres:
Closing in favor of #1876

PR #1904, opened 2026-08-26
Option to show both title and logo of items

tal-sarid:
## Description
Replaces the on/off “prefer logos” switch with three title display modes: Title only, Prefer logo only, and Prefer both. Existing off/on values migrates to Title only / Prefer logo only, so the default behavior does not change.

A logo-only header can be hard to read sometimes: many logos are stylized, low-contrast, or unrecognizable at a glance. Title only is always clear, but it loses the artwork that makes a title instantly recognizable. Prefer both keeps the readable name and the visual cue when a logo exists, and falls back to the title when it does not.

On item pages, Prefer both puts the title on the start side and the logo on the backdrop side (also RTL-aware).

The screensaver follows the same setting: both means logo in the top corner and title along the bottom; the other modes render title/logo at the bottom corner as before.

The player already shows both always and is unchanged. It is not following the new setting because when playing an episode of a tv show, the logo will be of the series itself so the title of the specific episode must be visible as well.

### Related issues
None.

### Testing
Tested on a physical Google TV Streamer 4K, checked manually relevant screens with various items.

## Screenshots
Screenshot_20260825_204752
Screenshot_20260825_204947
Screenshot_20260825_205620
Screenshot_20260825_213505

## AI or LLM usage
Implementation drafted with AI. I reviewed the code and cleaned it up manually afterwards while making sure minimal changes were done to achieve the goal and without any ugly code sections.


damontecres:
Thanks for the contribution, but I don't think I will merge this. This would another setting that needs to maintained and tested whenever a layout is changed. Unfortunately, I don't see a huge benefit.

Also, I held the same view about the logos being hard to read or contrast issues, etc, which is why logos were not included originally. But I've found most logos are great and it's easy to swap to a different logo as well.

tal-sarid:
> Thanks for the contribution, but I don't think I will merge this. This would another setting that needs to maintained and tested whenever a layout is changed. Unfortunately, I don't see a huge benefit.
> 
> Also, I held the same view about the logos being hard to read or contrast issues, etc, which is why logos were not included originally. But I've found most logos are great and it's easy to swap to a different logo as well.

My experience with the logos are different than yours and it is not that easy to change them either. Having both fixes everything without degrading the visuals of the UI (when using titles only mode provided).

Anyway, I see that we disagree about this but will you please reconsider merging this one if I re-open it as an experimental and unsupported feature instead?

I'll restore the original boolean option of course and will make sure to just add the logo element when the new feature is on instead of alternating existing ones like now. I'll implement it as unintrusive as possible with only code additions and when there is a layout change that this one is getting in the way - you can remove the feature as it is experimental and not a stable one. WDYT?

PR #1927, opened 2026-09-01
Add controls to edit playlists

damontecres:
## Description
Adds support to remove items from a playlist and move items within a playlist

The remove option is in the context menu. Moving has buttons on the right side. The buttons are available if the user has edit permissions and the playlist is sorted by "Default" (the playlist order). Remove from playlist option is available if the user has edit permissions.

Also, thumb images are now preferred instead of primary/poster images since the layout is much wider than tall.

### Related issues
Closes #1881

### Testing
Emulator with both a user with & without permissions to edit

### With move buttons
playlist_buttons Large

### Without move buttons
playlist_no_buttons Large

## AI or LLM usage
None

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/ui/components/ContextMenu.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/ui/components/ContextMenu.kt
@@ -41,6 +41,7 @@ sealed interface ContextMenu {
     data class ForBaseItem(
         val fromLongClick: Boolean,
         val item: BaseItem,
+        val index: Int = UNSPECIFIED_INDEX,
         val chosenStreams: ChosenStreams?,
         val showGoTo: Boolean,
         val showStreamChoices: Boolean,
@@ -73,6 +74,10 @@ sealed interface ContextMenu {
         val index: Int,
         val actions: QueueContextActions,
     ) : ContextMenu
+
+    companion object {
+        const val UNSPECIFIED_INDEX = -12
+    }
 }
 
 data class ContextMenuActions(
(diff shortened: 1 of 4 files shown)


schmitzkr:
Thanks for quickly addressing this!
I hit something while testing on the TV emulator:

With the cursor on playlist row 3 or lower, pressing left opens the nav drawer instead of moving to the Play/Shuffle controls. From row 1 it lands on Shuffle rather than Play.

Claude code told me that it looks like it's from commenting out the per-item focusProperties { left = playButtonFocusRequester } in PlaylistItems. The one on the LazyColumn doesn't reach the rows because the focusGroup() sits between them, so Compose falls back to geometric search and picks the nearest nav drawer icon.

Since each row now has the up/down/more buttons, maybe put that left property on the ListItem inside PlaylistItem only? That way the item still routes to Play but the buttons keep their natural left behavior.

Thanks again!

damontecres:
> I hit something while testing on the TV emulator:
> With the cursor on playlist row 3 or lower, pressing left opens the nav drawer instead of moving to the Play/Shuffle controls. From row 1 it lands on Shuffle rather than Play.

Thanks for testing it out! Yes, I commented out the `focusProperties` while I was reworking the layout and forgot to add it back.

Fixed in 59365aa1e085d27422e48eba23491277bd2173b7 by adding an `onExit` focus property to the entire list instead of individual list items.

schmitzkr:
> > I hit something while testing on the TV emulator:
> > With the cursor on playlist row 3 or lower, pressing left opens the nav drawer instead of moving to the Play/Shuffle controls. From row 1 it lands on Shuffle rather than Play.
> 
> Thanks for testing it out! Yes, I commented out the `focusProperties` while I was reworking the layout and forgot to add it back.
> 
> Fixed in [59365aa](https://github.com/damontecres/Wholphin/commit/59365aa1e085d27422e48eba23491277bd2173b7) by adding an `onExit` focus property to the entire list instead of individual list items.

I tested this quickly and it behaves as expected.

PR #1950, opened 2026-09-09
Fix person credits: movie release dates, full filmography, sorted order

### LINKED ISSUE #1873: Person page: movie credits lose their year, and the filmography stops at 25 cast + 25 crew
### Description

On a person page opened from the library, the Discover row of TMDB credits has two problems.
The first is a data mapping defect; the second is an inconsistency with the other person page
in the same app.

**1. Movies in that row never show a year.**

Cameron Diaz, opened from *Charlie's Angels*: the row holds 28 cards, and exactly **two** of
them show a year. Both are TV series. Every film shows an empty line where the year goes.

The card prints the year from `releaseDate`:

```kotlin
// ui/cards/DiscoverItemCard.kt:195
text = item?.releaseDate?.year?.toString() ?: "",
```

and for credits that field is filled from `firstAirDate` alone:

```kotlin
// services/SeerrService.kt:427 — createDiscoverItem(CreditCast), and :454 for CreditCrew
releaseDate = toLocalDate(credit.firstAirDate),
```

TMDB puts a movie's date in `releaseDate` and a series' date in `firstAirDate`, so films come
out with no date at all — which is exactly the 26-of-28 above. Both fields are already in the
schema (`CreditCast` and `CreditCrew` in `app/src/main/seerr/seerr-api.yml`), and the mapping
right next door, for search results, already handles it:

```kotlin
// services/SeerrService.kt:410 — createDiscoverItem(SearchResult)
releaseDate = toLocalDate(search.releaseDate ?: search.firstAirDate),
```

So the fix is the same expression in the two credit mappings:

```diff
-                releaseDate = toLocalDate(credit.firstAirDate),
+                releaseDate = toLocalDate(credit.releaseDate ?: credit.firstAirDate),
```

**2. The filmography is capped at 25 + 25, and the cap is not applied on the other person page.**

Those 28 cards are 25 acting credits plus 3 crew credits. TMDB lists **132** acting credits
and 3 crew credits for her, so the row drops 107 titles:

```kotlin
// services/SeerrService.kt:130-140
val cast = credits.cast?.take(25)?.map { createDiscoverItem(it) }.orEmpty()
val crew = credits.crew?.take(25)?.map { createDiscoverItem(it) }.orEmpty()
cast + crew
```

`DiscoverPersonPage` calls the same endpoint with no `take` at all, so the same actor reached
through Discover shows a full filmography while the one reached from the library shows at most
50 — and the row has no "see all" affordance to reach the rest.

If the cap exists to keep the row cheap, a `take` after sorting would at least cut the least
relevant entries rather than whatever order the API happened to return.

**3. While the file is open: the order.**

`cast + crew` concatenates two lists, so the row is not in any order a viewer can read — and
the concatenation also discards the order Seerr itself sends (it returns both lists by TMDB
`voteCount`, descending). Once the dates above are correct, sorting by date descending is a
one-liner in both places, and de-duplicating by `id` removes the double entries for people who
are both cast and crew on the same title.

I have kept this in the bug report because it is the same three lines of code, but if you would
rather have it as an Ideas discussion, say the word and I will move it there.

### Reproduction steps

1. Configure Seerr/Jellyseerr in Wholphin.
2. Open a movie in the library, e.g. *Charlie's Angels*, and open one of its cast members
   (Cameron Diaz) from the cast row.
3. Scroll to the Discover row on the person page.
4. Count the cards: 28. Note that only the two TV series show a year, and that the acting
   credits stop at 25.

### Experimental settings

No

### App Version

v1.0.6

### Server Version

10.11.11

### Device

NVIDIA Shield TV Pro 2019

SergeiAK:
## Description

Person pages that get a filmography from Seerr had three problems, all in the same place.

The release date was taken from `firstAirDate` alone — the field TMDB fills for series — so
movie cards showed no year at all. The library person page capped the row at 25 cast + 25 crew while
`DiscoverPersonPage` had no cap, because each screen carried its own copy of the same logic.
And `cast + crew` left the row in no order a viewer can read, with a title appearing twice when
a person is both cast and crew on it.

This PR:

* maps credit dates as `releaseDate ?: firstAirDate` — the expression the search result mapping
  already uses;
* moves the credits logic into a single `SeerrService.personCredits()` used by both screens, so
  they cannot drift apart again;
* drops the 25 + 25 cap, de-duplicates by TMDB id and sorts by release date descending, with
  undated entries last.

Rows on movie and series pages are untouched: only the `BaseItemKind.PERSON` branch of
`similar()` changed.

Two things worth your opinion:

* Without the cap, the row for a person with a long filmography is long — 135 cards for Cameron
  Diaz, for example. If you would rather keep a cap, the row could keep the newest 25 and gain a
  `ViewMoreCard` opening the full grid — `Destination.DiscoveredItem` with a `PERSON` item
  already routes to `DiscoverPersonPage`. I am happy to add that in a second commit.
* The sort is fixed to release date descending. It now lives in one function, so a different key
  or a user-facing choice would be cheap to add later.

### Related issues

Fixes #1873

### Testing

CI in my fork ran the repository's own PR workflow on this branch: `pre-commit` (ktlint) and
`assembleDefaultDebug` with `testDefaultDebugUnitTest` are both green.

I then installed a debug build of the same code on an NVIDIA Shield TV Pro 2019 (Android 11),
against Jellyfin 10.11.11 with Jellyseerr, and opened Cameron Diaz from the cast row of
*Charlie's Angels* before and after the change:

* movie cards show a year now; before, the only two cards in the row with a year were the two
  TV entries;
* the row is ordered by release date descending, movies and series interleaved, instead of the
  API order of all cast followed by all crew;
* the row is no longer cut at 25 + 25 — it now runs back to her 1952 credit, and the entries
  TMDB has no date for sit at the very end of the row.

Not tested on any other device; nothing in the change is device specific.

## Screenshots

Person page of Cameron Diaz, reached from the cast row of *Charlie's Angels*. The library rows
above the Discover row are the same in every shot and are useful as a reference: they always had
their years.

Before, start of the Discover row — no year under any card:
pr1873-1-before-start

Before, end of the Discover row — the row stops here, and the only two years in it belong to the
two TV entries:
pr1873-2-before-end

After, start of the row — years on the movie cards, ordered by date descending:
pr1873-3-after-start

After, end of the row — the list now reaches 1952, with the undated entries last:
pr1873-4-after-end

## AI or LLM usage

The code was written with Claude Opus 5 (Claude Code): the two mapping lines, the extracted
`personCredits()` function and the two call sites. I reviewed every line and can explain it.
This description was written by me in Russian and translated into English with the same
assistant.

### DIFF
--- a/app/src/main/java/com/github/damontecres/wholphin/services/SeerrService.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/services/SeerrService.kt
@@ -124,21 +124,7 @@ class SeerrService
                             }
 
                             BaseItemKind.PERSON -> {
-                                api.personApi
-                                    .personPersonIdCombinedCreditsGet(personId = it)
-                                    .let { credits ->
-                                        val cast =
-                                            credits.cast
-                                                ?.take(25)
-                                                ?.map { createDiscoverItem(it) }
-                                                .orEmpty()
-                                        val crew =
-                                            credits.crew
-                                                ?.take(25)
-                                                ?.map { createDiscoverItem(it) }
-                                                .orEmpty()
-                                        cast + crew
-                                    }
+                                personCredits(it)
                             }
 
                             else -> {
@@ -149,6 +135,24 @@ class SeerrService
             } else {
                 null
             }
+
+        /**
+         * Get the combined movie & TV credits of a person as [DiscoverItem]s
+         *
+         * The two lists returned by the API are merged, de-duplicated by id since a person can be
+         * both cast and crew on the same title, and sorted by release date descending with undated
+         * entries last
+         */
+        suspend fun personCredits(personId: Int): List<DiscoverItem> =
+            api.personApi
+                .personPersonIdCombinedCreditsGet(personId = personId)
+                .let { credits ->
+                    val cast = credits.cast?.map { createDiscoverItem(it) }.orEmpty()
+                    val crew = credits.crew?.map { createDiscoverItem(it) }.orEmpty()
+                    (cast + crew)
+                        .distinctBy { it.id }
+                        .sortedByDescending { it.releaseDate }
+                }
 
         suspend fun getTvSeries(item: BaseItem): TvDetails? =
             if (active.first()) {
@@ -424,7 +428,7 @@ class SeerrService
                 availability =
                     SeerrAvailability.from(credit.mediaInfo?.status)
                         ?: SeerrAvailability.UNKNOWN,
-                releaseDate = toLocalDate(credit.firstAirDate),
+                releaseDate = toLocalDate(credit.releaseDate ?: credit.firstAirDate),
                 posterUrl =
                     createImageUrl(
                         ImageType.PRIMARY,
@@ -451,7 +455,7 @@ class SeerrService
                 availability =
                     SeerrAvailability.from(credit.mediaInfo?.status)
                         ?: SeerrAvailability.UNKNOWN,
-                releaseDate = toLocalDate(credit.firstAirDate),
+                releaseDate = toLocalDate(credit.releaseDate ?: credit.firstAirDate),
                 posterUrl =
                     createImageUrl(
                         ImageType.PRIMARY,
--- a/app/src/main/java/com/github/damontecres/wholphin/services/SeerrService.kt
+++ b/app/src/main/java/com/github/damontecres/wholphin/services/SeerrService.kt
@@ -154,6 +154,32 @@ class SeerrService
                         .sortedByDescending { it.releaseDate }
                 }
 
+        /**
+         * Create a [DiscoverItem] for a person in the library so that the full grid of their
+         * credits can be opened from their page
+         *
+         * @return the item or null if the person has no TMDB id
+         */
+        fun discoverPerson(item: BaseItem): DiscoverItem? =
+            item.data.providerIds
+                ?.get("Tmdb")
+                ?.toIntOrNull()
+                ?.let {
+                    DiscoverItem(
+                        id = it,
+                        type = SeerrItemType.PERSON,
+                        title = item.name,
+                        subtitle = null,
+                        overview = item.data.overview,
+                        availability = SeerrAvailability.UNKNOWN,
+                        releaseDate = null,
+                        posterUrl = null,
+                        backDropUrl = null,
+                        logoUrl = null,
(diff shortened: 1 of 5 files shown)


damontecres (changes requested):
This looks pretty good! The sort changes make a lot of sense.

> Without the cap, the row for a person with a long filmography is long [..] the row could keep the newest 25 and gain a ViewMoreCard

I think this would be good. I think once there's more than 25-30 items in a row, it becomes difficult to navigate, so switching to the grid with "View More" is better. You could use either a static 25 or reuse the max items in home row setting.

SergeiAK:
Added in a second commit.

The row on the library person page now stops at the same limit as the home page rows and ends with a `ViewMoreCard`; the card opens `DiscoverPersonPage`, which keeps showing the whole filmography as a grid. Of the two options you offered I took the setting rather than a static 25: **Max items on home page rows** already defaults to 25, so out of the box the row behaves exactly as you described, and a user who has shortened their home rows gets the shorter row here too. Turning it into a constant is a one-line change if you would rather not tie the two together.

The only new piece is `SeerrService.discoverPerson()`, which builds the `DiscoverItem` for the person from their TMDB id, since `Destination.DiscoveredItem` needs one to route to the grid. The sort and the de-duplication stay in `personCredits()`, so the row and the grid cannot disagree about the order.

### Testing

CI in my fork ran `pre-commit` (ktlint), `assembleDefaultDebug` and `testDefaultDebugUnitTest` on this branch: all green. I then installed a debug build of the same code on an NVIDIA Shield TV Pro 2019 (Android 11) against Jellyfin 10.11.11 with Jellyseerr, and opened Cameron Diaz from the cast row of *Charlie's Angels* again.

End of the Discover row — 25 cards, then the "View more" card:

 
pr1950-2-row-end

The grid behind it — her full filmography, 133 entries after duplicates are merged, in the same order as the row:


pr1950-3-grid

Coming back from the grid returns to the person page with the focus restored on the "View more" card rather than at the start of the row:


pr1950-4-back

### AI or LLM usage

As with the first commit: the code was written with Claude Opus 5 (Claude Code) and reviewed by me line by line; this comment was written by me in Russian and translated with the same assistant.

damontecres (approved):
This looks good, thanks!

I pushed a small change to pass the view more index so that the grid starts at the next item instead of the first.

PR #1962, opened 2026-09-14
Update intent parsing to allow for user switching to home page

damontecres:
## Description
Makes the `itemId` parameter optional allowing just passing `serverId`/`userId` to switch the user and load the app at the home page for that user.

Dev note: I added a new assert function `assertIs` which asserts that an object `is` of a type. It also has a contract so the compiler will smart-cast usages after the `assertIs`. Checking a result type is a very common pattern in the unit tests, so this assert will simplify a lot of current & future tests.

### Related issues
Closes https://github.com/damontecres/Wholphin/discussions/1824

### Testing
Emulator and added unit tests for `parseIntent`

## Screenshots
N/A

## AI or LLM usage
None


Kooldude93:
Would this be a suitable place to request an example or two of the URI format for the user switching to be added in the intents documentation or would you prefer I open a new issue?

Edit: I've done some testing with the debug build and found that the following URI request works fine. Feel free to add it into your documentation.

`adb shell am start -d 'wholphin://view?serverId=da8674e5GR9647f7a9a18ffa482004eb'`

`adb shell am start -d 'wholphin://view?serverId= da8674e292b647f7a5r23ffa482004eb\&userId=r7ho26d75b1441179f0f47ba8293ea11'`

PR #1964, opened 2026-09-15
Guard against crash if episode has `seasonId` but no `seriesId`

damontecres:
## Description
When clicking on an episode, the app assumes if there is a season ID for the episode, then there must be a series ID. However, on my JF 12.0 server, an episode (with no metadata) appeared on my Continue Watching and it has a season ID (which in turn has a series ID) but no series ID.

I'm not sure if this really a valid situation or if it's a server bug, but Wholphin shouldn't crash at least. So this PR checks both `seasonId` and `seriesId` and if either is null, falls back to the generic "Episode Page". Similarly, if a season has no `seriesId` (again, not sure if a valid situation), a crash won't occur, however there is no generic "Season Page", so an error page is shown.

### Related issues
TODO, check JF server issues, maybe fixed in 12.1?

### Testing
Emulator

## Screenshots
N/a

## AI or LLM usage
None


damontecres:
A few hours later and the episode was fixed, so this was probably just an edge case during a library scan. But guarding against a potential crash is still a good idea.

PR #1968, opened 2026-09-15
Fix NoCompatibleStream when playing a fresh in-progress DVR recording

mackandall:
## What

Pressing Play on a DVR recording that's still actively being written does nothing — no error, no toast — if that particular recording has never been played before. Logcat shows:

```
PlaybackViewModel$changeStreams: Error in PostedPlaybackInfo: NoCompatibleStream
```

...after which the app just navigates back to the series page with no indication anything went wrong.

## Root cause

`play()` only treats `BaseItemKind.TV_CHANNEL` as needing no explicit `mediaSourceId`. A `Recording` item goes through `streamChoiceService.chooseSource(base, itemPlayback)`, which picks from the item's *cached* `MediaSources` (from a plain item `GET`) — whose `Id` is the item's own guid (`MediaSourceManager.GetStaticMediaSources` server-side). That id gets sent as an explicit `mediaSourceId` in the `PlaybackInfo` request.

But for an *actively recording* item, the server resolves the real source dynamically via `MediaSourceManager.GetRecordingStreamMediaSources`, whose `Id` is the active recording's own id — never the item's guid. The server's `MediaInfoHelper.ResolvePlaybackMediaSources` filters by the requested `mediaSourceId`, finds nothing, and returns `ErrorCode: NoCompatibleStream`.

This only reproduces on a *fresh* first-ever play attempt for a given in-progress recording — a previously-played one can take a different path and work fine, which is probably why it hadn't surfaced before.

## Fix

Treat an in-progress recording (`item.type == BaseItemKind.RECORDING && dto.status == "InProgress"`) the same as a live TV channel in `play()` — skip static source selection, pass `sourceId = null`, let the server resolve fresh.

Verified on-device against two independent fresh in-progress recordings on two different shows/channels: before the fix, instant `NoCompatibleStream`; after, a normal `Transcode` playback decision and the video plays immediately.

## Separate observation (not addressed by this PR)

Even without this specific root cause, `changeStreams` silently swallows `response.errorCode` instead of surfacing it to the user (a toast/snackbar) — worth considering separately, since any future `NoCompatibleStream` case will otherwise look like "nothing happens" again.


damontecres:
Thanks for submitting this PR. Please fix the PR description to match the template provided when opening a pull request. Additionally, please review the [AI/LLM policy](https://github.com/damontecres/Wholphin/blob/main/CONTRIBUTING.md#policy-for-ai-or-llms) again.

mackandall:
Apologies. Will fix with human language and resubmit.

PR #1969, opened 2026-09-15
Fix opening a movie DVR recording's details page

mackandall:
## What

Selecting a DVR recording of a *movie* (e.g. from the Recorded Movies grid) shows a bare "Unsupported item type: Recording" screen instead of the movie's details page. No crash, just a dead-end text screen.

This equally affects an *episode* recording opened via its own standalone details destination — it just doesn't surface as often, since browsing an episode inline within its season page uses a different code path that renders its own header directly.

## Root cause

`BaseItem.destination()` converts a `BaseItem` into a navigation `Destination` via `when (type) { ... }`. A DVR recording's own `type` is *always* `BaseItemKind.RECORDING`, regardless of whether the underlying content is a movie or a series episode — there was no `RECORDING` case, so it fell into `else -> Destination.MediaItem(this)`, which carries `type = RECORDING` straight through. `DestinationContent.kt`'s own dispatch `when (destination.type)` has cases for `SERIES`, `MOVIE`, `EPISODE`, etc., but none for `RECORDING` either — hence the "Unsupported item type" fallback branch and its literal message.

## Fix

Added a `BaseItemKind.RECORDING ->` branch in `destination()`, mirroring the existing `EPISODE` branch's logic: check `data.seasonId` — if present (a recorded episode), redirect to `Destination.SeriesOverview(...)` with the right season/episode ids; otherwise (a recorded movie), redirect to `Destination.MediaItem(itemId = id, type = BaseItemKind.MOVIE, ...)`.

This is safe because `MovieDetails`/`MovieViewModel` only use `destination.itemId` to re-fetch the item fresh from the server — the `type` field is purely used for the top-level page-dispatch `when`, not by the destination page itself.

Verified on-device against an actively-recording movie: the details page now renders correctly (with a "Recording" indicator from a separate change), and pressing Play starts playback normally.


damontecres:
Thanks for submitting this PR. Please fix the PR description to match the template provided when opening a pull request. Additionally, please review the [AI/LLM policy](https://github.com/damontecres/Wholphin/blob/main/CONTRIBUTING.md#policy-for-ai-or-llms) again.

